PR #71, opened 2020-02-26
front end changes by marty

martyboggs:
I've only ever done pull requests on BitBucket.... but I think did this right
@tylerthome


tylerthome:
looks good

PR #142, opened 2020-05-02
questions with material UI

martyboggs:
Still working on the state changes.


martyboggs:
navigate to /profile to answer some questions

app/src/pages/Register.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react"
```
arshiamasih, line 1: unclear on what this component is @martyboggs @tylerthome - do we need this or can we delete it?

app/src/pages/ProfileEdit.tsx
```diff
@@ -0,0 +1,126 @@
+import * as React from "react"
```
arshiamasih, line 1: hi i know this seems minor but is profile edit a confusing name for this? this isn't necessarily part of an edit flow once a profile has been built. just want to avoid confusion! @martyboggs @tylerthome

app/src/pages/ProfileEdit.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { Button, LinearProgress } from '@material-ui/core';
+import { Question } from "../models/Question"
+
+const Container = styled.div`
+  margin: 30px auto;
+  padding: 0 15px;
+  max-width: 1140px;
+`
+
+const StyledButton = styled(Button)`
```
arshiamasih, line 14: looks like we are rebuilding a button in a lot of place, let's at some point think about button reusability! we can build a generic button that take an onclick handler as a prop and passes in a callback. and reuse it throughout the app.

nothing to change here - throwing it out for us to think about with this and other components we reuse often.

app/src/components/ProfileEdit/QuestionField.tsx
```diff
@@ -0,0 +1,78 @@
...
+                question.options && question.options.map((option: any) => 
+                  <FormControlLabel key={option.value} value={option.value} control={<Radio />} label={option.label} />
+                )
+              }
+            </RadioGroup>
+          </FormControl>
+        }
+        {
+          question.type === 'checkbox' &&
+          <FormControl component="fieldset">
```
arshiamasih, line 51: we should build the MUI components out as generic components so we can reuse them through the app. i think we can thinking about this in a separate task / PR but wanted to mention it.

app/src/components/ProfileEdit/QuestionField.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from "react"
+import { useParams, useHistory } from "react-router"
+import { useHostHomeData } from "../../data/data-context"
+import { ProgressPlugin } from "webpack"
+import { Question } from "../../models/Question"
+import styled from "styled-components"
+import { FormControl, FormControlLabel, FormGroup, RadioGroup, Radio, Checkbox, TextField } from '@material-ui/core';
```
arshiamasih, line 7: these are all the components that would likely make up our mini 'library' - just an fyi

app/src/pages/Register.tsx
arshiamasih, line 1: unclear on what this component is @martyboggs @tylerthome - do we need this or can we delete it?

app/src/pages/ProfileEdit.tsx
```diff
@@ -0,0 +1,126 @@
+import * as React from "react"
+import QuestionField from "../components/ProfileEdit/QuestionField"
```
arshiamasih, line 2: hi i know this seems minor but is profile edit a confusing name for this? this isn't necessarily part of an edit flow once a profile has been created. just want to avoid confusion! @martyboggs @tylerthome

app/src/components/ProfileEdit/QuestionField.tsx
martyboggs: Sure, it's always good to not repeat oneself. NRO = not repeat oneself

app/src/pages/ProfileEdit.tsx
martyboggs: I thought this page could double as a "register new user" and "profile edit" page. Once all questions are answered the first time, they could all be displayed at once for the user to edit and keep up-to-date. This was before Tyler brought up the difference between matching questions and user personal information questions though. We'll need to discuss the flow more to get this right.

app/src/pages/Register.tsx
martyboggs: Oh yeah, I built out some routes to get an idea of the flow with organizations and host/guest. I'll delete this now.

PR #162, opened 2020-05-20
QuestionPage component

martyboggs:
I thought it would be good to build a reusable wrapper QuestionPage component that uses the Question component.

It will receive an array of questions with validation and conditional logic info and whether to display all questions or one at a time... yeah


tylerthome:
Solid idea, Marty. I like that you have added this in a way that we can 'opt in' when the requirements are finalized.

arshiamasih:
hey Marty - great idea! my 2 cents here would be that we should be explicitly clear on what entails a parent-child relationship (i see this in the follow up questions for example). also, we have expanded on some needs which we will likely need to reflect in our Question interface -- for example a completion status. i have some initial thoughts but will leave this up to you to decide on the best approach!

martyboggs:
I'll reopen this pull request on the master branch

PR #164, opened 2020-05-22
verify user etc

HassenH1:


app/src/App.tsx
```diff
@@ -23,132 +22,144 @@ import { Guest } from './models'
...
+    logout() //something has to be updated on auth0 side to go to the right route
+
   }
 
-  const { isInitializing, isAuthenticated, user, logout } = useAuth0();
+  React.useEffect(() => {
+    const fetch = async () => {
+      if (isAuthenticated) {
+        let fetch = new Fetcher<Host>('checkEmail')
+        let data: any | undefined = await fetch.getByEmail(user)
```
tylerthome, line 72: This can be made more concise in a future refactor. Since we already used the email to query the back end, and the back end determined whether this email had a registered account, we can just use the appropriate HTTP status code to indicate whether the account was found or not. A response payload only needs to be read if an existing account was found (HTTP 20*), which provides the account ID and type for subsequent routing to user's dashboard

app/src/App.tsx
```diff
@@ -23,132 +22,144 @@ import { Guest } from './models'
...
+                    />
+                    <Route
+                      path="/guests/:id"
+                      component={GuestProfilePage}
+                    />
+                    <Route
+                      path="/admin/hosts"
+                      component={AllHosts}
+                    />
+                    <Route
```
tylerthome, line 149: As we have discussed, this may change depending on our business requirements for registration. It is possible that our user will not have an account ID until completing a minimum step from the registration process. In this case, we will refactor the following route to exclude the ID param

app/src/data/ApiWrapper.ts
```diff
@@ -58,6 +78,10 @@ export class Fetcher<T> {
             JSON.stringify(item)
         )) as string
     }
+
```
tylerthome, line 81: This is correctly implemented, well done. Let's discuss whether this ultimately belongs here at this level of abstraction: Fetcher is currently a generic wrapper for a RESTful interface supporting Get, GetById, Put, Delete, Post. Adding a getByEmail() method could lead its users to trying that method on resources for which it does not apply, e.g. HostQuestion.

hosthome.py
```diff
@@ -122,7 +122,18 @@ def get_element_by_id(self, collection_name, id):
 
         self._log('get_collection', 'item = {}'.format(item))
         return item
+    ###########################################################attempt
```
tylerthome, line 125: Previously discussed with back end team: we are leaving this hard-coded value temporarily to test our authentication+registration flow. This will be made dynamic

app/src/App.tsx
HassenH1: you're right about using status code rather than using the email itself

PR #166, opened 2020-05-27
Reg flow component library

arshiamasih:
Built a few MUI input components for the registration flow, to be: 
1. rendered based on the question type
2. used to create some of the more custom forms (contact, personal info etc), rendered on question type 'custom-form'

**Components:**
- checkbox (should be used for multi select)
- radio buttons 
- text input 
- large text input
- select dropdown 

these are all mounted on: localhost:8080/componentlibrary via the AllInputs.tsx file for testing and viewing purposes.
 
**Props:**
Each of the components take explicit MUI props. These can be modified to take any props as needed.

**Styling:**
i have started on a themeprovider as well to be scoped to this part of the app, but will continue styling on a separate PR.


app/src/components/Registration/Dropdown.tsx
```diff
@@ -6,54 +6,59 @@ import {
...
+interface Props {
     value: string
     onChange: (event: object) => void
-    placeholder: string
     id: string
-    itemValue: string
-    className: string
-    htmlFor: string
     helperText: string
     options: Array<string>
```
tylerthome, line 26: We may want to support a differing 'value' and 'text' property for the dropdown options. For example, our generic "select one" question types, the application should submit the ID of the response, which will be different from its display text

app/src/components/Registration/Dropdown.tsx
```diff
@@ -6,54 +6,59 @@ import {
...
                             name,
                             id,
                         }}
                     >
+                        <MenuItem value="">
+                            <em>None</em>
+                        </MenuItem>
                         {options.map((menuItem: string) => {
                             return (
-                                <MenuItem value={itemValue}>
```
tylerthome, line 50: this might be where an ID might be mapped into the value prop instead of the display text

app/src/components/Registration/Checkbox.tsx
```diff
@@ -8,19 +8,21 @@ import {
...
+
+//checkbox for multi-select
+const CheckboxInput = (props: Props) => {
+    const { onChange, label, options, helperText } = props
+    let checked: boolean
 
     return (
         <>
-            <form className={className}>
+            <form>
```
martyboggs, line 25: Epic Components Arshia!
We'll have to decide if we need <form> elements when we tackle validation. Since they can't be nested, it would be tricky to conditionally include them based on whether a parent component has already included one.

arshiamasih: great point @martyboggs! those can definitely be removed. nice catch.

PR #172, opened 2020-05-30
question groups

martyboggs:


martyboggs:
I also changed setAnswer to get the question by index rather than searching by id.

I think editing the text fields is broken. I'll set this as a draft till it's fixed.

tylerthome:
Should resolve issue #174

PR #181, opened 2020-06-05
Create Showstoppers for questions that were answered no

HassenH1:


HassenH1:
all the changes I made on App.tsx was to get it to work without requiring a user
also, I made a profile button on nav to help navigate to the questions

app/src/components/ProfileEdit/QuestionPage.tsx
```diff
@@ -52,122 +76,170 @@ export const QuestionPage = (props: Props) => {
         if (!groups[groupI][subgroupI]) groups[groupI][subgroupI] = []
         groups[groupI][subgroupI].push(state.questions[i])
     }
-
+    
     function setAnswer(index: number, answer: any) {
-        var state2 = {...state}
+
+        const indexQuesitons = [0, 1, 2, 3, 4, 5, 6, 7, 8] //for modalv
```
martyboggs, line 82: Good job, Hassen. The modal is working well.
You can make it more dynamic by adding a property to the QuestionType interface. That way it will work for other question sets we pass in as JSON. The property can be called showstopper with the value being the answer as a string.

arshiamasih:
Hey hassen - awesome work! Can you pull the modal out as a separate component similarly as we have done with the component library? the styling should follow the same format as well as this ties back to our theming. 

marty makes some great points about ensuring the showstopper logic is working dynamically. we need some indication or way of knowing on each question whether there will be a showstopper or not.

i understand we are in 'mock' mode since we don't have all the questions but let's also think of all the scenarios that might trigger the showstopper based on the information we do have.

app/src/components/ProfileEdit/QuestionPage.tsx
```diff
@@ -52,122 +76,170 @@ export const QuestionPage = (props: Props) => {
...
+            const testing = indexQuesitons.includes(index)
+            if(testing){
+                setDisableSubmit(true)
+                setOpen(true)
+            } 
+        } else {
+            setDisableSubmit(false)
+        }
+        //for modal^
+        var state2 = { ...state }
```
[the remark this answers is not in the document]
martyboggs: That's my bad. I was trying to do this in a single expression without using map, but I don't think it's possible. I'll fix it in the my next PR.
```
var state2 = {...state}
state2.questions[index].answer = answer
setState(state2)
```

HassenH1:
okay everything is good to go now

app/src/components/ProfileEdit/QuestionPage.tsx
```diff
@@ -52,122 +56,136 @@ export const QuestionPage = (props: Props) => {
...
         groups[groupI][subgroupI].push(state.questions[i])
     }
-
+    
     function setAnswer(index: number, answer: any) {
-        var state2 = {...state}
+        var state2 = { ...state }
         state2.questions[index].answer = answer
+
+        if(answer === "no" && state2.questions[index].showstopper){
```
martyboggs, line 64: Almost there... Can you make this `if (answer === state2.questions[index].showstopper) {`
Then make the questions with showstoppers have `showstopper: 'no'`
That way, we can have a showstopper for any answer.

Then make sure to resolve the merge conflicts with master.

PR #197, opened 2020-06-13
Confirmation modal

HassenH1:
I just need to fix the styling

### DIFF
--- a/app/src/components/MUIModal/ConfirmationModal/ConfirmationModal.tsx
+++ b/app/src/components/MUIModal/ConfirmationModal/ConfirmationModal.tsx
@@ -25,7 +25,7 @@ function ConfirmationModal(props: ConfirmationModalProps) {
     return (
         <MUIModal handleClose={onCancel} modalOpen={modalOpen}>
             <h2 id="transition-modal-title" style={{ color: 'red' }}>
-                Confirmation Modal
+                Confirmation Modal Example
             </h2>
             <p>{confirmationText}</p>
             <Button
--- a/app/src/components/MUIModal/ConfirmationModal/ConfirmationModal.tsx
+++ b/app/src/components/MUIModal/ConfirmationModal/ConfirmationModal.tsx
@@ -1,41 +1,48 @@
 import React from 'react'
 import MUIModal from '../MUIModal'
-import { styles } from './styles' //taking this out for now
+import { styles } from './styles'
 import Button from '@material-ui/core/Button'
 import { withStyles, WithStyles } from '@material-ui/core/'
 
-// interface ConfirmationModalProps { //taking this out for now
 interface ConfirmationModalProps extends WithStyles<typeof styles> {
+    confirmationText: string
     handleClose: () => void
     modalOpen: boolean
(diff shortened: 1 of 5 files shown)


app/src/components/MUIModal/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import MUIModal from '../MUIModal'
+import { styles } from './styles'
+import Button from '@material-ui/core/Button'
+import { withStyles, WithStyles } from '@material-ui/core/'
+
+interface ConfirmationModalProps extends WithStyles<typeof styles> {
+    confirmationText: string
+    handleClose: () => void
```
tylerthome, line 9: We can scrap this handleClose() from the top-level props interface for this component. This is a lower-level concern to be handled internally by this component. We will discuss whether we want the user to be able to click outside of the modal to close without explicitly providing a response, but if we _do_ want to support that, onCancel will suffice to pass as a callback to MuiModal's handleClose prop

app/src/components/MUIModal/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+        modalOpen,
+        handleClose,
+        onConfirm,
+        onCancel,
+        confirmationText,
+        classes,
+    } = props
+
+    return (
+        <MUIModal handleClose={onCancel} modalOpen={modalOpen}>
```
tylerthome, line 26: This takes care of my comment above, and makes the top-level handleClose prop obsolete

app/src/components/MUIModal/MessageModal/MessageModal.tsx
```diff
@@ -0,0 +1,21 @@
...
+    modalHeadingText: string
+    modalMessage: string
+}
+
+function MessageModal(props: MessageModalProps) {
+    const { handleClose, modalOpen, modalHeadingText, modalMessage } = props
+    return (
+        <MUIModal handleClose={handleClose} modalOpen={modalOpen}>
+            <h2 style={{ color: 'red' }}>{modalHeadingText}</h2>
+            <p>{modalMessage}</p>
```
tylerthome, line 16: Somewhere around here should live a Button element allowing the user to acknowledge the Message, instead of forcing them to click outside the modal window or on the 'X' to make it disappear. We may want to keep the top-level handleClose() callback handy here since there is no conditional behavior associated with closing this modal; let's bring this up with UX

tylerthome:
Now that @martyboggs has approved, and we resolved one of the issues in the review by @arshiamasih during our team discussion on 6/13 we should be ready to merge this code. @HassenH1 please add a brief overview description to this PR detailing what is included in the changes being merged.

HassenH1:
this PR is basically two subclasses created using the MUIModal component, a ConfirmationModal, and a generic MessageModal. When a developer needs a quick modal they can either just call one of these two components and pass in the necessary props needed or they can create a custom modal using the MUIModal component.

PR #199, opened 2020-06-17
Checkboxes fix

martyboggs:


martyboggs:
I added the updated gender question from Xio's design. I need to change the checkboxes component a little to get it working. Works with nested questions now too.

PR #201, opened 2020-06-23
Context lite

arshiamasih:
sets up an initial and bare bones pass at state management for the host dashboard using Context.

the purpose here is not optimal code (yet), but rather to check that the pattern is set up correctly and is sound. 

for now/this PR, the focus is on retrieving the host registration questions via context. i have made some light edits to the api wrapper that we will revise later as we expand for additional business logic. 

the QuestionPage/Question components still use props from ProfileEdit, as that is using a hard coded questions array, instead of the back end data. however we can now eliminate the questions prop drilling. 

*note: have not addressed enums yet.


app/src/data/host-context.tsx
```diff
@@ -0,0 +1,125 @@
...
+            'useHostDashboardData must be used within a HostDashboardProvider'
+        )
+    }
+
+    const [data, dispatch] = context as [
+        HostDashboardData,
+        React.Dispatch<HostDashboardAction>
+    ]
+
+    return {
```
tylerthome, line 121: Let's discuss what abstractions will be the most useful for driving state from lower-level components. Ideally, a consumer does not need to understand the action and payload types to use this hook via the `dispatch` function and we can provide more descriptive callbacks which encapsulate the use of dispatch.

app/src/data/host-context.tsx
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function hostDashboardReducer(
+    state: HostDashboardData,
+    action: HostDashboardAction
+): HostDashboardData {
+    switch (action.type) {
+        case HostDashboardActionType.BeginFetchQuestions: {
+            return {
+                ...state,
```
tylerthome, line 49: Since this case is for a well-defined action type, we can simplify by omitting the `payload` property from the call. In other words, `action.type === BeginFetchQuestions` already provides all the necessary information to determine what message should be displayed in the Loader. Unless you foresee that this text could vary for different dispatch calls for this action, we can alleviate the caller from needing to pass this string every time.

app/src/data/host-context.tsx
```diff
@@ -0,0 +1,125 @@
...
+                loaderState: {
+                    loading: true,
+                    message: action.payload as string,
+                },
+            }
+        }
+        case HostDashboardActionType.FinishFetchQuestions: {
+            return {
+                ...state,
+                loaderState: {
```
tylerthome, line 59: For this case, the text "Finished Loading" will never display, since this modal will not be visible as soon as this state change is propagated to the UI. One thing we could add here (during dev/debug ONLY) would be some descriptive message to indicate that this loader's was not properly assigned. Alternatively, eliminating line 61 is acceptable.

arshiamasih: oops i did not realize this was for actually displaying the text! good point.

app/src/data/host-context.tsx
arshiamasih: yes absolutely. i did not expand any further because I still feel unclear on business logic.

arshiamasih:
going to move forward with merging this so we can build upon it.

PR #226, opened 2020-08-18
Host bio

HassenH1:


app/src/components/HostFormDetail/HostFormGender.tsx
```diff
@@ -0,0 +1,288 @@
+import React, { useState } from 'react'
```
tylerthome, line 1: @HassenH1 for the JSX elements using identical style attributes please define these variables of type `CSSProperties` 

For example, one repeated rule set apparent in this snippet could be factored out into the following reusable object:
```
const fieldContainerStyles: CSSProperties = {
   display: 'flex',
   flexDirection: 'row',
   alignItems: 'center',
   fontSize: '18px',
}
```

and used in the JSX as `<div style={fieldContainerStyles}></div>` where applicable

HassenH1: Just added this

HassenH1:
Actually, Dont merge this
I forgot to add something

HassenH1:
Ready to Go

Issue #217, opened 2020-08-08
Extend ApiWrapper to support the submission of responses for User Information and generic registration questions


tylerthome:
### Overview
Add methods to the ApiWrapper to POST/PUT responses for all parts of the registration process

### Action Items
- [ ] POST a guest information object to `/api/guest`
- [ ] POST a host information object to `/api/host`
- [ ] POST a generic response array MatchingQuestion or ShowstopperQuestion to `/api/questions/<:questionId>`

Issue #216, opened 2020-08-08
Extend ApiWrapper to support fetching MatchingQuestion, QualifyingQuestion


tylerthome:
### Overview
Add ApiWrapper logic to HTTP fetch, JSON parse and return TS objects for the newly-created MatchingQuestion and ShowstopperQuestion interfaces

### Action Items
- [ ] Add `getShowstopperQuestions` to return `Array<ShowstopperQuestion>`
- [ ] Add `getMatchingQuestions` to return `Array<MatchingQuestion >`

### Resources
* Study the corresponding Flask API endpoint to appropriately call and handle the logic at the HTTP layer

ExperimentsInHonesty:
@tylerthome does this need to be in the UAT column?

PR #230, opened 2020-08-22
Host context 

arshiamasih:
Addresses issues: #216 and #217 

Except:
POST a host information object to /api/host

^This language is vague - unclear if this is in reference to the initial profile object to which all answers are stored (when the user selects their profile type), or for specific host registration forms. 

If it is the former it should be handled in the ProfileSelection and I can address it in a separate ticket. 

Please provide explicit feedback on what needs to be corrected.


app/src/data/host-context.tsx
```diff
@@ -1,14 +1,16 @@
 import * as React from 'react'
-import { QuestionType } from '../models/QuestionType'
-import { ResponseValue } from '../models/ResponseValue'
+import { MatchingQuestionType } from '../models/MatchingQuestionType'
+import { ShowstopperQuestionType } from '../models/ShowstopperQuestionType'
+import { HostResponse } from '../models/HostResponse'
 import { ApiWrapper } from './ApiWrapper'
 
 const HostDashboardContext = React.createContext({})
 const hostsFetcher = new ApiWrapper()
```
tylerthome, line 9: For constant-time lookup of questions by ID, these will ideally be modeled as `Map<number, ShowstopperQuestionType>` etc. this will require one-time iteration through the question array to build this structure as soon as the data is loaded from the server. Add exported callbacks for `getShowstopperQuestionById(id:number)` . This change will correspond to refactored UI components and integer ID'd questions from the API

Issue #221, opened 2020-08-11
Home Picture Uploader


tylerthome:
### Overview
We need to have a method to upload photos to be displayed on the Host profile.

### Action Items
Integrate the file upload component into a view for for the Home Photos for the Host Registration flow
Make sure the new component is able to:

- [x] Upload and store multiple images in component state
- [x] Matches the wireframe diagram (linked below)
- [x] Has access to the API Wrapper module for uploading up confirmation by user

### Resources/Instructions

Create the view specified in [this wireframe diagram](https://www.figma.com/file/5uXe3GeE99l16uBf5hucWb/Home-Unite-Us?node-id=542%3A180), using the new file uploader component

![image](https://user-images.githubusercontent.com/9952154/89951840-feab0700-dbe0-11ea-8ad7-d0219798f234.png)

abbyhipp:
Has been completed per #251

PR #231, opened 2020-08-25
Upload image component regarding issue #221

HassenH1:
Upload image component good to go
Just need the API wrapper stuff and we are all set

### DIFF
--- a/app/src/components/HostFormDetail/HostFormInfo.tsx
+++ b/app/src/components/HostFormDetail/HostFormInfo.tsx
@@ -208,7 +208,7 @@ const HostFormInfo: React.FC = () => {
                                         justifyContent: 'flex-start',
                                     }}
                                 >
-                                    <Button>
+                                    <Button style={{ color: '#55B1EB' }}>
                                         <span
                                             style={{
                                                 padding:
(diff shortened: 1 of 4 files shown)


app/src/components/HostFormDetail/HostFormInfo.tsx
```diff
@@ -224,7 +224,11 @@ const HostFormInfo: React.FC = () => {
                                     <Button
                                         variant="contained"
                                         color="primary"
-                                        style={{ margin: '0 3rem 0 7rem' }}
+                                        style={{
```
tylerthome, line 227: In the near future these styling concerns should be informed by the theme provider, or at least factored into a local definition rather than inlined

app/src/components/UploadImage/UploadImageButton.tsx
```diff
@@ -33,17 +46,131 @@ function UploadImageButton() {
...
+                        <div style={ContainerStyles}>
+                            {selectedImage[i] ? (
+                                <div style={{ width: '100%', height: '100%' }}>
+                                    <img
+                                        src={selectedImage[i]}
+                                        width="100%"
+                                        height="100%"
+                                    />
+                                </div>
+                            ) : (
```
tylerthome, line 100: rather than invoking `addMoreButtons()` call it as a JSX component (should be renamed to a capitalized component name, e.g. `<AddMoreButtons />`

HassenH1: just resolved this issue

Issue #234, opened 2020-08-30
ApiWrapper support for matching and qualifying question submissions by ID


tylerthome:
### Overview
ApiWrapper should allow the submission of matching and qualifying questions by ID so that the application can support dynamic sets of these question types

### Action Items

- [x] Add method to API wrapper to submit a `PUT` request to the route `/api/host/registration/qualifying/<questionId>`
- [x] Add method to API wrapper to submit a `PUT` request to the route `/api/host/registration/matching/<questionId>` 

### Resources/Instructions
The `body` of the HTTP request should have the following shape:
```javascript
{
    "email": string;
    "response": string | Array<number> | number
}
```

For now, you may consider the "response" property to have `any` type as we validate the persistence and retrieval for the Dashboard and/or Confirmation views

tylerthome:
Note that the ApiWrapper currently supports previous design accepting any generic question response by ID, this enhancement will allow the storage of these responses as separate subcollections in the data store, as well as possible deviations in models for these questions in the future

PR #235, opened 2020-09-02
ApiWrapper methods PUT by id: qualifying & matching questions

arshiamasih:
Addresses issue #234 

**Splits the generic putHostResponse method into two methods:**
- putMatchingResponse
- putShowstopperResponse

**Updates in files:**
- ApiWrapper
- host-context
- MatchingQuestion
- ShowstopperQuestion

Note: The proposed model is commented out for now as the current `HostResponse` model is in use in the data-context. A separate model could be created.
```
{ 
  email: string
  responseValues: string | Array<number> | number
}
```
Disregard any commented out code unrelated to issue #234

### DIFF
--- a/app/src/data/ApiWrapper.ts
+++ b/app/src/data/ApiWrapper.ts
@@ -106,7 +106,7 @@ export class ApiWrapper {
     private hostShowstopperResponse: Fetcher<string>
     private hostMatchingResponse: Fetcher<string>
 
-    public constructor(id?: number | string, email?: string) {
+    public constructor(id?: number | string) {
         this.guestFetcher = new Fetcher<Guest>('guests')
         this.hostShowstopperQuestionsFetcher = new Fetcher<
             ShowstopperQuestionType
@@ -126,10 +126,10 @@ export class ApiWrapper {
             `/api/v1/hostRegisterQuestions/${id}`
         )
         this.hostShowstopperResponse = new Fetcher<string>(
-            `/api/v1/hostRegisterQuestions/${email}`
+            `/api/host/registration/qualifying/${id}`
         )
         this.hostMatchingResponse = new Fetcher<string>(
-            `/api/v1/hostRegisterQuestions/${email}`
+            `/api/host/registration/matching/${id}`
         )
     }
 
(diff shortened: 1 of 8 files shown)


app/src/data/ApiWrapper.ts
```diff
@@ -123,6 +125,12 @@ export class ApiWrapper {
         this.hostQuestionsResponse = new Fetcher<string>(
             `/api/v1/hostRegisterQuestions/${id}`
         )
+        this.hostShowstopperResponse = new Fetcher<string>(
```
tylerthome, line 128: Please remove the "/api" prefix from these URLs, as that value is prepended by `Fetcher`'s constructor to potentially support other environments and API versions.

tylerthome (changes requested):
Looks good so far! Aside form the requested change, this is ready to merge.  The use of `Promise.all` will yield better concurrency, and these resource types are similar enough that destructuring the 2D array in the reducer seems reasonable, but may merit some explanatory comments near that logic..

arshiamasih:
@tyler any suggestions on a better approach to this? I used the 'any' type on the payload to follow the approach I used, and I don't know how to resolve or improve that either.
_and these resource types are similar enough that destructuring the 2D array in the reducer seems reasonable, but may merit some explanatory comments near that logic.._

arshiamasih:
the `HostResponse ` interface is being used in data-context in what I believe is hard-coded matching results. 

Should I for now create a separate model for the qualifying and matching host response so as not to potentially break anything?

Also where will be the user email be accessible from?

tylerthome:
> @tylerthome any suggestions on a better approach to this? I used the 'any' type on the payload to follow the approach I used, and I don't know how to resolve or improve that either.

The simplest way to keep the current pattern while avoiding the use of offset indexing for the destructure would be to assign directly into named collections, i.e. `const [myShowstopperQuestions, myMatchingQuestions] = Promise.all([...])` and instead of modeling the `action.payload` as an array, use an interface with these collections as named properties. Then, the reducer can query each by name without typecasting from `any` or using integer indices.

> the HostResponse interface is being used in data-context in what I believe is hard-coded matching results.
> 
> Should I for now create a separate model for the qualifying and matching host response so as not to potentially break anything?

Yes, good idea, and we will create a new issue to deprecate all obsolete models from the proof-of-concept implementation.

> Also where will be the user email be accessible from?

Since this context provider is a child of the `Auth0Context.Provider>` the `useAuth0` hook will be available within both `<HostDashboardContext.Provider>` as well as its client components communicating via the hook.

PR #246, opened 2020-09-23
stepwise questions connecting api

martyboggs:
No new features, just getting this component to load again


tylerthome (approved):
Looks great, Marty! With these being retrieved from the context provider at this level, as opposed to props, the transition to route-based initialization will be much more convenient.

Issue #244, opened 2020-09-20
JSON schema for custom themes


tylerthome:
### Overview
To support a custom themes per customer organization, define a JSON schema describing all customizable properties of a theme

### Action Items
- [x] Create an example JSON file specifying customer theme

### Resources/Instructions
The schema should include at least the following information:
* Filename or URL of logo
* Primary color
* Name of customer/organization

tylerthome:
This issue to also include as a consideration the capability for the Material UI theme provider to initialize its primary color to the value from the customer settings object

tylerthome:
Resolved as of PR #247

PR #247, opened 2020-09-23
Model for Account Settings (top level theme)

arshiamasih:
Addresses issue #244 

Defines a TS interface describing all customizable properties of an account level theme.

The schema includes at least the following information:
- Filename or URL of logo
- Primary color
- Name of customer/organization
- Account ID

Regarding the capability for the Material UI theme provider to initialize its primary color to the value from the customer settings object:
If we are using a top level account context, we could create a hook / expose the data to the mui theme.tsx (where the mui theme is configured) or alternatively create a wrapper ('adapter') component.


app/src/models/AccountSettings.ts
```diff
@@ -0,0 +1,10 @@
+/* schema describing the customizable properties of the theme */
+/* first defined at account onboarding */
+
+export interface AccountSettings {
+    accountId: string
+    accountName: string
```
tylerthome, line 6: this field may become an reference by ID rather than URI to support dev/qa/prod environments, pending discussion of user story for modifying this attribute. We will defer that discussion to post-MVP

tylerthome (approved):
Looks great, Arshia! This will define the contract for creation/update of this resource between the Home Unite Us application and webservice, as well as with calendaring service API

PR #251, opened 2020-10-03
238

HassenH1:


app/src/data/ApiWrapper.ts
```diff
@@ -198,6 +200,123 @@ export class ApiWrapper {
         return await this.hostLanguageForm.putResponse(item)
     }
 
+    // TODO: Hassen
```
tylerthome, line 203: Better practice here is to use `Promise<void>` so we explicitly call out that no return value is expected

app/src/data/ApiWrapper.ts
```diff
@@ -198,6 +200,123 @@ export class ApiWrapper {
...
+                body: payload,
+            })
+            if (response.status !== 200) {
+                throw new Error(`file ApiWrapper: ${response.statusText}`)
+            }
+        } catch (e) {
+            throw new Error(`getHostPictures got an ${e}`)
+        }
+    }
+
```
tylerthome, line 225: Let's convert this one to return `Promise<void>` as well

tylerthome (changes requested):

tylerthome:
@HassenH1  - I requested a couple one-line changes, otherwise this looks ready to merge

tylerthome:
Good work, Hassen!

PR #255, opened 2020-10-12
250

HassenH1:


tylerthome (approved):
Good start on this blueprint, let's unify our error handling to support flexibility on client side (via model extensibility)

PR #256, opened 2020-10-21
245

HassenH1:


HassenH1:
I just realized there is a Mistake, so please do not approve this yet

blueprints/case.py
```diff
@@ -0,0 +1,77 @@
...
+                          template_folder='templates')
+
+client = pymongo.MongoClient()    
+db = client[DB_NAME]
+collection = db['case'] 
+case_repository = Case_Repository(collection, db)
+
+@case_api.route('/create_case', methods=['POST'])
+def create_case():
+  try:
```
tylerthome, line 20: This variable name makes sense since it is the "response" provided by the user with the case information, but this could be confusing in this context because we are looking at the HTTP _request_ and the HTTP response is what we send back to the client. This name could be made more descriptive with a noun like "payload" , "request_body" or something even more specific like "case_data"

blueprints/case.py
```diff
@@ -0,0 +1,77 @@
...
+
+  try:
+    response = request.json
+
+    if response is None:
+      return jsonify("bad request, nothing in response in update_case_status"), 400
+
+    if not response['case_id'] or not response['status_id']:
+      return jsonify("Both fields must be filled"), 400
+
```
tylerthome, line 48: A more descriptive name for this variable would be something like "update_successful" or "is_acknowledged" depending on what is being returned by this function call

tylerthome (approved):
Looks good @HassenH1 , please check out the notes for future reference. We will get this merged so that front-end dev team can begin to query these endpoints.

Issue #352, opened 2021-10-22
Create profile selection with routing scaffold for Host, Guest and Coordinator home views


tylerthome:
### Overview
Using the Figma design describing the initial application flow, create the React components to support selection of Host, Admin or Guest such that the appropriate view is rendered.

### Action Items
- [ ] Import relevant SVG markup into project
- [ ] Templatize foreground and background color for state management via props
- [ ] Change Tabs menu within Coordinator Dashboard to appear as child items of the Home menu option of left panel navigation bar

### Resources/Instructions
* Figma URL for Profile type selection screen: https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/Current-HUU-Everything-Figma?node-id=1230%3A3611
* Figma URL for Coordinator Dashboard: https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/Current-HUU-Everything-Figma?node-id=1355%3A1

PR #355, opened 2021-11-03
#352 / Create profile selection with routing scaffold for Host, Guest and Coordinator home views

erikguntner:
This pull request includes:

- Add baseline global CSS resets by including the CSSBaseline component from Material UI
- Update the header image to use the SPY logo
- Create and style header Icon Buttons
- Create and style the home page by adding the headers and buttons to select which role to sign in as
- Import SVG icons from FIgma and turned them into JSX components. This allows better control over adding and removing styles
- Convert dashboard tabs to children navigation displayed in the side nav bar
- Add logout functionality to profile menu
- Replace loading image with circular progress indicator and center content
- Refactoring some of the current layout components by cleaning up HTML


tylerthome (approved):
These changes look good @erikguntner  let's continue with this pattern for wrapping our SVGs as you proposed

PR #392, opened 2022-04-06
Redux Toolkit and Redux Toolkit Query integration

erikguntner:
This contains the integration of Redux Toolkit into the app architecture. That includes:

- Adding Redux Toolkit as a dependency to the app
- Wrapping the app in a Redux Provider
- Creating an example counter state slice (to be removed in the future)
- Used Redux Toolkit code generation to generate api slice based off Swagger specification


app/src/services/api.ts
```diff
@@ -0,0 +1,156 @@
+import {emptySplitApi as api} from './emptyApi';
```
tylerthome, line 1: This looks promising, let's run with this for now. Additionally, we should see how the generator does on a spec that uses all of the features we need around auth and potentially websocket connections to see if we'll need to fill any gaps.

tylerthome (approved):

Issue #400, opened 2022-04-30
Add OpenAPI decalarations or HTTP routes for model stubs


tylerthome:
### Overview
Add schemas to the OpenAPI YAML specification file for leaf-level models defined in new sqlalchemy declarations

### Action Items
Add the `components`  `schemas` and  `paths` entries to the OpenAPI yaml file for the following classes:

- [x] HousingProgramServiceProvider
- [ ] IntakeQuestion
- [ ] IntakeResponseValue

### Resources/Instructions

* SQLAlchemy models:
https://github.com/hackforla/HomeUniteUs/blob/main/api/openapi_server/models/database.py

* OpenAPI spec for Home Unite Us:
https://github.com/hackforla/HomeUniteUs/blob/main/api/openapi_server/openapi/openapi.yaml

PR #402, opened 2022-05-10
feat: add schemas for new models and implement CRUD endpoints for HousingProgramServiceProvider

shubha-rajan:
Changes in this PR
- Modularize openapi spec file for readablility
- Update Dockerfile to bundle modularized yaml files
- Add schemas for all new classes
- Add paths for HousingProgramServiceProvider
- Generate files (some of these generated files might not be needed.
- Basic implementation of controller methods for HousingProgramServiceProvider

Towards #400, I'll do the paths and controller for the other classes in separate PRs to keep the PR size manageable


tylerthome (approved):
Looks great! Getting the auth modules incorporated into the yaml spec, adding the swagger gen build step and an initial integration between our http and db layers are major bonuses here

PR #414, opened 2022-07-02
remove sb and regenerate deps from latest

tylerthome:


tylerthome:
Storybook w/ Vite seems to have this bug that we're encountering: https://github.com/storybookjs/builder-vite/issues/248

We may be able to just remove dependency for `@storybook/addon-interactions` however we aren't expressly declaring that dependency, it's getting pulled during install of `@storybook/builder-vite` and I assume would get re-installed on every clean workstation setup.

This issue causes our installation to pull in webpack libraries and 20+ high severity vulnerabilities are reintroduced to the codebase

PR #434, opened 2022-10-19
Front and back end implementation for Hosts

erikguntner:
This PR acts as a reference for the creation of a data model, controller, and serializer on the backend. As well as, the generation and utilization of data fetching hooks on the front end.

This PR contains:
- The creation of a Host data model in SQLAlchemy
- The creation of a serializer and deserializer for said model
- The creation of a host controller with get and post routes
- Addition of routes to OpenAPI spec
- Generation of Redux data fetching hooks in services folder
- Adding cache invalidation tags to the host service for the automated refetching and updating of related data
- Creating a hosts page that shows the use of query and mutation hooks to fetch and update data

I added ts-node as a development dependency since it was required by the `@rtk-query/codegen-openapi` package


tylerthome (changes requested):
Looks great! This should be good once we handle the argument and exceptions in the controller implementations

PR #468, opened 2023-02-23
Saves user to database

emeldar:
If using the regular sign up method, save user and return error if uniqueness constraint violated.
If using token method, only add user if doesn't exist. 
Adds email column to user table.


erikguntner:
Awesome! Looks good

PR #471, opened 2023-02-25
Layout component

erikguntner:
This PR creates a separate layout component from the new header and then wraps all existing routes. This will need to be updated when we have different layouts for different parts of the application.


jed-stewart (commented):
@erikguntner This looks good

PR #482, opened 2023-03-14
Cleans up backend server

emeldar:
Removes outdated yaml file
Removes unnecessary auto-generated OpenAPI model files 
Adds alembic for database migration tracking
Removes local database from git


api/alembic/env.py
```diff
@@ -0,0 +1,86 @@
...
+
+from openapi_server.models import database as db
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+if config.config_file_name is not None:
```
jed-stewart, line 21: What if there is not a config?

jed-stewart: or it is not found, maybe we should exception

emeldar: This is specifically the alembic config, this env.py code isn't ever loaded when we're running the app normally, it's only loaded when making migrations using alembic. And if no alembic.ini file exists, it will error out for other reasons

jed-stewart (approved):

erikguntner (approved):

PR #485, opened 2023-03-22
Forgot Password flow

erikguntner:
This PR sets up the views and logic for the forgot password flow and includes changes such as:

- Created a form context that wraps the forgot password screens and collects the necessary data as the user moves through the form
- Views for the forgot password, verification code, and reset password pages. Along with validation, loading states, and error handling
- A code input component that allows you to type in a 6-digit code and corresponding tests
- Removes old auth0 components
- Renames reset password to confirm forgot password to make it more in line with the AWS commands and to not get it confused with an endpoint for setting a new password during guest account creation
- A responsive container component for the forms that adhere to the grids laid out in the designs
- A password field component with an end icon that shows and hides the visibility of passwords

This PR was getting long so I left a few things that still need to be done:
- Write end-to-end tests to at least test the happy path
- The final form submission relies on data being collected throughout the process, and right now we don't handle what happens if that data is lost at different steps. This can be due to the user navigating away, refreshing the page, etc.
- Moving the buttons to the bottom on mobile devices
- Resend code logic


erikguntner:
https://user-images.githubusercontent.com/27253583/226777612-8262d6ab-d119-4f24-9957-29bfee9dec6e.mov

PR #491, opened 2023-04-14
Password Validity Component

### LINKED ISSUE #479: Password strength indicator
### Overview
Currently, password validation on the signup screen is displayed using text but needs to be switched out for [this](https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/HUU-Everything-Figma-Nov-2022?node-id=5357%3A69697&t=L6Z9YpJjw5oGIu9Z-4) component in Figma. It's currently only displayed in two states so you may have to follow up with the design team for more information on functionality and appearance in different states.

The current validation schema for Cognito is:
- Minimum length of 8
- Requires numbers
- Require special character
- Require uppercase letters
- Require lowercase letters

### Action Items
- Create a component that lists the strength requirements
- Each requirement can be in a neutral, successful, or error state and displays the corresponding icon
- Add the component to the signup form

mira-kine:
This PR consists of a reusable password validation component using Yup and Formik. It is currently used in the `SignUp` view for QA testing and viewing.

Closes #479 : Password Strength Indicator

Here is a screen recording of what we have right now: 

https://user-images.githubusercontent.com/90009901/231942711-efc75305-fdd3-4e14-94db-13b0738a4b8c.mov

What it does:

- Lists strength requirements and provides neutral or check according to what user is typing in
- Disabled Submit button until all fields are valid (including email)
- Is added to sign up form

I didn't make the pull from main yet to incorporate the ability to see the password yet, but that would be a better way to test it after this PR.


app/src/components/common/PasswordValidationSchema.tsx
erikguntner, line 1: I think this could be moved to utils folder or something similar. It might be a better fit there since it's not a component

app/src/components/authentication/SignUpForm.tsx
```diff
@@ -16,44 +16,46 @@ import CloseIcon from '@mui/icons-material/Close';
...
+
+      onSubmit: values => {
+        onSubmit(values);
+      },
+    });
+
+  // set errors to "contains" state, passdown to password validation as prop
+  // and dynamically change UI according to what is in errors array
+
+  const [errorsLeft, setErrorsLeft] = useState<string[]>([]);
```
erikguntner, line 52: I'm wondering if there is a way we could encapsulate this state and the validation logic inside the PasswordValidation component since we have to use this component in multiple places. Maybe if we just pass the password from Formik down and handle the validation and state update in a useEffect?

app/src/components/common/PasswordValidationSchema.tsx
```diff
@@ -0,0 +1,40 @@
...
+    })
+    .matches(/^(?=.*[a-z])/, {
+      message: {lowercase: 'Must contain at least one lowercase character'},
+    })
+    .matches(/^(?=.*[A-Z])/, {
+      message: {uppercase: 'Must contain at least one uppercase character'},
+    })
+    .matches(/^(?=.*[!@#%&])/, {
+      message: {
+        special: 'password must contain at least one special character',
```
jed-stewart, line 18: Can we change line 18 to "password must contain these special characters (!@#%&)"

jed-stewart (changes requested):

mira-kine:
Fixed all type issues and moved errors state to password validation component in common. Also moved schema into utils.

PR #498, opened 2023-05-05
Update Cognito Client URLs

erikguntner:
This PR updates urls on the backend to use the new Cognito Client. It also updates the logic to disable the sign up button with an invalid form input.


api/openapi_server/controllers/auth_controller.py
```diff
@@ -217,7 +217,7 @@ def token():
     client_id = COGNITO_CLIENT_ID
     client_secret = COGNITO_CLIENT_SECRET
     callback_uri = request.args['callback_uri']
-    cognito_app_url = 'https://homeuudemo.auth.us-east-1.amazoncognito.com'
+    cognito_app_url = 'https://homeuniteus.auth.us-east-1.amazoncognito.com'
```
ju1es, line 220: sweet. couple questions as I'm still familiarizing myself with the code base,
- should we put this here or at the top with the other env variables?
- unit test? safety net that `requests.post(token_url, auth=auth, data=params)` on `233` is called with the updated url.

app/src/components/authentication/SignUpForm.tsx
```diff
@@ -31,14 +31,13 @@ export const SignUpForm = ({
   errorMessage,
   setErrorMessage,
 }: SignUpFormProps) => {
-  const {handleSubmit, handleChange, values, touched, errors, isValid} =
+  const {handleSubmit, handleChange, values, touched, errors, isValid, dirty} =
```
ju1es, line 34: nit but also not familiar with `formik`, what's the difference between `isValid` and `dirty`? are they both related to invalid form input?

ju1es (approved):

ju1es:
pre-approving, just a couple questions

app/src/components/authentication/SignUpForm.tsx
mira-kine: isValid returns true if there are no errors, and dirty means that the user has touched the field aka it is not deeply equal to the initial values

mira-kine (approved):
Looks good, and thank you for cleaning up line 87 in sign up form!

app/src/components/authentication/SignUpForm.tsx
erikguntner: Yep! Mira summed it up. I was running into an issue with the disabled submit button and found [this](https://formik.org/docs/migrating-v2#isvalid) in the Formik docs.

api/openapi_server/controllers/auth_controller.py
erikguntner: Both are good ideas. I'll go ahead and it to the top. Maybe we want to make this an environment variable in the future? I don't think it's considered a secret, but I could see us potentially wanting to have dev/prod clients based on the environment. I'm not sure what our unit testing setup is like on the backend right now, but I like that idea as well.

PR #511, opened 2023-05-26
Forgot password error message

erikguntner:
This PR includes the following updates:

- Displays error messages to the user in the case of data loss during the forgot password process
- Wrote e2e happy path test for forgot password flow
- Updated testing library versions
- Added a Github action to run Cypress tests
- Added the msw library to mock network requests in integration testing


app/src/views/ForgotPasswordCode.tsx
```diff
@@ -26,6 +32,18 @@ export const ForgotPasswordCode = () => {
     setFieldValue('code', code);
   };
 
+  if (!email) {
+    return (
+      <FormContainer>
```
tylerthome, line 37: In a future changeset we may be able to factor this out so that it is reused here and in the `ResetPassword` component

tylerthome (approved):
Looks good!

PR #521, opened 2023-06-13
Add Debugging to README

ju1es:
## Problem
Unable to step through/debug requests to api.

## Solution
- Add "how-to" in readme
- Disable reloader.

## Resources
- [debugger was stalling and this was the fix](https://stackoverflow.com/questions/25504149/why-does-running-the-flask-dev-server-run-itself-twice).
- [known subprocess debugging issue](https://github.com/microsoft/vscode-python/issues/3473).


api/openapi_server/__main__.py
```diff
@@ -24,7 +24,7 @@ def main():
                 arguments={'title': 'Home Unite Us'},
                 pythonic_params=True)             
     app.add_error_handler(AuthError, handle_auth_error)
-    app.run(port=8080, debug=True)
+    app.run(port=8080, debug=True, use_reloader=False)
```
ju1es, line 27: When this isn't enabled, I can't get the vscode debugger to start flask app. I suspect it's the debugger that's unable to attach to the subprocess created by Werkzeug. I included some background info in the resources but I think it's only to have functionality of making updates to the app while it's live. If we're on ci/cd, I'm assuming we'd have to re-deploy anyway? As for local development, I'm thinking that restarting the server with your new changes is a small price for being able to debug requests. Open to any thoughts here.

erikguntner:
I'll check out the branch and play around with it a bit. So basically the hot reloading feature in Flask interferes with the VS Code debugger?

ju1es:
@erikguntner exactly yep

ju1es:
closing this as I'm adding the toggling in a separate [pr](https://github.com/hackforla/HomeUniteUs/pull/527)

ju1es:
@erikguntner updating only the readme as I merged the changes to toggle the 'use_reloader' config in a separate pr

paulespinosa:
@ju1es @erikguntner When using an external debugger, the [Flask external debugging documentation](https://flask.palletsprojects.com/en/2.3.x/debugging/#external-debuggers) suggests to disable the built-in debugger and reloader:

```python
app.run(debug=True, use_debugger=False, use_reloader=False)
```

Keep in mind that this `app.run()` function is only used for local development: [connextion API doc](https://connexion.readthedocs.io/en/stable/autoapi/connexion/apps/flask_app/index.html#connexion.apps.flask_app.FlaskApp.run) and [flask API doc](https://flask.palletsprojects.com/en/2.3.x/api/#flask.Flask.run)

ju1es:
@paulespinosa ah! thank you for digging into that. I'm a little confused though. When we debug, we set our breakpoints, set flask's reloader and debugger to false, then start the api via the following command?
```
python3 -m openapi_server
```

paulespinosa:
That looks right.

Issue #503, opened 2023-05-10
Make Cognito redirect URLs configurable in web service


tylerthome:
### Overview
Update references to the redirect URLs for Cognito in the Flask application

### Action Items
* Establish a configuration convention for the Flask application, e.g. use a json or xml config format, env file
* Create the configuration for the redirect URI for HUU tenant in AWS Cognito
* Reference the configured value in all statements requiring the redirect URL

### Resources/Instructions
The redirect URL tells Cognito where to send a user after authentication. The HomeUniteUs API is deployed in a variety of environments, e.g. `dev-ec2`, `test-ec2`, `prod-ec2`, `host-local`, `docker-local`.  Each of these may have different URLs where the web service is hosted, which the app will need to specify in requests to Cognito for user management operations like token validation.

PR #524, opened 2023-06-19
Cognito redirect ur ls configurable

stevbark:
Should fix the redirect issue when we update the URL value in AWS.


api/.env.example
```diff
@@ -5,4 +5,5 @@ COGNITO_REDIRECT_URI=http://localhost:4040/signin
 COGNITO_USER_POOL_ID=
 SECRET_KEY=
 COGNITO_ACCESS_ID=
-COGNITO_ACCESS_KEY=
\ No newline at end of file
+COGNITO_ACCESS_KEY=
+URL=http://localhost:4040
```
paulespinosa, line 9: A more descriptive name would help me understand the purpose of this variable.

paulespinosa:
This change is for Issue #503, right?

stevbark:
> This change is for Issue #503, right?

yes. https://github.com/hackforla/HomeUniteUs/issues/503

api/.env.example
ju1es: +1

ju1es:
sorry in advance if I'm not understanding this correctly! but what's the purpose of `.env.example`? if it's the `.env` that will run in production, i think we'd specify environment variables in the docker file right? I started cooking up a config system that'll generate at runtime,

https://github.com/hackforla/HomeUniteUs/pull/527/files

maybe we add this there?

paulespinosa:
@ju1es 
> but what's the purpose of `.env.example` ?

My understanding is that it's a file that we can check into version control (git) to demonstrate the environment variables that need to be set within an deployment environment without the actual values (secrets). 

> i think we'd specify environment variables in the docker file right?

If the `.env` file exists when building the docker image, it will be pulled into the docker image and used by the application when it starts up. I imagine the prod version of the `.env` file is stored somewhere in AWS that the continuous integration build can access. The CI process is something I'm not familiar with yet, though.

stevbark:
New changes pushed. more descriptive name added.

PR #536, opened 2023-07-12
Global nav

erikguntner:
This PR includes the following updates to the global nav bar:

- Adding the Settings, Help, and Logout items to the dropdown menu
- Creating the settings page
- Display user initials in Avatar

We aren't currently collecting the user's name so right now it uses a fake name.

### DIFF
--- a/app/src/components/common/Header.tsx
+++ b/app/src/components/common/Header.tsx
@@ -24,7 +24,10 @@ const userName = 'Hank Hill';
 
 function getInitials(name: string) {
   const splitName = name.split(' ');
-  return splitName[0][0] + splitName[1][0];
+
+  if (splitName.length === 1) return splitName[0][0];
+
+  return splitName[0][0] + splitName[splitName.length - 1][0];
 }
 
 interface Props {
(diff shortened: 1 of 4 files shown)


erikguntner:
https://github.com/hackforla/HomeUniteUs/assets/27253583/3dfb5c26-6b63-45d2-9779-364639b6ea0a

app/src/components/common/Header.tsx
```diff
@@ -18,7 +18,14 @@ import {useAuth} from '../../app/hooks/useAuth';
...
 import {useAppDispatch} from '../../app/hooks/store';
 import {setCredentials} from '../../app/authSlice';
-import {useNavigate} from 'react-router-dom';
+import {Link, useNavigate} from 'react-router-dom';
+
+const userName = 'Hank Hill';
+
+function getInitials(name: string) {
+  const splitName = name.split(' ');
+  return splitName[0][0] + splitName[1][0];
```
paulespinosa, line 27: This should have more checks. For example, check if `splitName` has more than 1 element.

paulespinosa (changes requested):

app/src/components/common/Header.tsx
erikguntner: Thanks! Added a check for length and also to use the last element in case it's not just first and last name.

paulespinosa (approved):
Thank you

PR #537, opened 2023-07-15
fix broken host config

ju1es:
## Problem
Flask app isn't accessible when started via Docker.

## Solution
- Bind to all server interfaces.
- Fix tests.

## Resources
[listen on all interfaces](https://forums.docker.com/t/docker-is-running-but-i-cannot-access-localhost-flask-application/82193)
[difference between 127.0.0.1 and 0.0.0.0](https://www.baeldung.com/linux/difference-ip-address)

### DIFF
--- a/api/configs/configs.py
+++ b/api/configs/configs.py
@@ -6,7 +6,7 @@ Never put any sensitive information here i.e. credentials, secrets, etc.
 '''
 
 class Config():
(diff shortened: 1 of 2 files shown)


erikguntner (approved):
Looks good!

paulespinosa (approved):

Issue #404, opened 2022-05-17
Section 1 | SubFlow 1a: Coordinator Account Creation


rossbb:
### The Problem:

Currently, Coordinators do not have a way to create an account for Home Unite Us. As a result, Coordinators cannot access Home Unite Usto view and/or manage Guest and Host applications. 

#### User Story:

As a Coordinator, I want to login to Home Unite Us so that I can track and manage the Guests and Hosts in the system.

### High Level Approach:

Enable Coordinators to create a Home Unite Us user account.

### Considerations:
* In the future (not this phase), the ability for Coordinators to add other Coordinators may be needed

### Goals & Success:

Success is if Coordinators can create a user account from the Home Unite Us website.

### Key Features: 

* Button to create a new account on the Home Unite Us webpage
* Option to select Coordinator (other option: Host)
* Field to enter an email address, or continue with Google or Apple
* Fields to input and confirm password, with validation
* A verification email that is sent to the inputted email address, containing a verification link

### Acceptance Criteria:

* Users can select create new account from the dev Home Unite Us homepage (dev.homeunite.us)
* Users can select user type (Coordinator, Host)
* Users can enter an email address or continue with Google, Apple
* Users can create a password
   * If the password meets the validation requirements, a verification email is sent to the provided email containing a verification link
   * If the password does not meet the validation requirements, the user will see which validation requirements were not met and can try again with another password
* Users will click the verification link in the email which will confirm the user account and redirect the user to the login portal
* The user can then enter an email address and password or continue with Google or Apple
   * If the login was successful, users will be redirected to the Coordinator home page
   * If the login was not successful, the user will see an error message and can try logging in again 
* User flows above can be completed on desktop, mobile, and tablet 

### Designs:

- [x] Review existing designs on Figma
- [x] Edit existing designs as needed
- [x] Designs are edited and ready for engineering. Link to the Figma: (share Figma link here when ready)

### Key Decisions:

* After successfully logging in for the first time, the user will be redirected to the app (versus the default redirect to an AWS page). The rationale is that users will have a better first experience with the app if redirected to the app versus the AWS page. This will require additional dev effort - if significant dev effort is needed will revisit.

edela0015:
https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/HUU-Everything-Figma-Nov-2022?node-id=6764-95988&t=CsDDFP3o1U1ubwPm-0

edela0015:
[...]

Issue #403, opened 2022-05-17
Section 1 | Subflow 1d: Host Account Creation


rossbb:
### The Problem:

Currently, Hosts do not have a way to create an account for Home Unite Us. As a result, Hosts cannot login to create a profile, a required step to host Guests on the platform.

#### User Story:

As a potential Host, I want to create an account on Home Unite Us so that I can login and apply to be a Host.

### High Level Approach:

Allow Hosts to create a Home Unite Us user account.

### Goals & Success:

Success is if Hosts can create a user account from the Home Unite Us website.

### Key Features:

* A button to create a new account from the Home Unite Us homepage
* Option to select Host (other option: Coordinator)
* Field to enter an email address or continue with Google or Apple
* Fields to input and confirm password, with validation
* A verification email that is sent to the inputted email address, containing a verification link

### Acceptance Criteria:

* Users can select create new account from the Home Unite Us homepage
* Users can select user type (Coordinator, Host)
* Users can enter an email address or continue with Google, Apple
* Users can create a password
    * If the password meets the validation requirements, a verification email is sent to the provided email containing a verification link
    * If the password does not meet the validation requirements, the user will see which validation requirements were not met and can try again with another password
* Users will click the verification link in the email which will confirm the user account and redirect the user to the login portal
* The user can then enter an email address and password or continue with Google or Apple
   * If the login was successful, users will be redirected to the Coordinator home page
   * If the login was not successful, the user will see an error message and can try logging in again
* The user flow can be completed across desktop, mobile, and tablet

### Designs:

- [ ]  Review existing designs on Figma
- [ ]  Edit existing designs as needed and share a link to the updated designs in this issue

### Key Decisions:

* After successfully logging in for the first time, the user will be redirected to the app (versus the default redirect to an AWS page). The rationale is that users will have a better first experience with the app if redirected to the app versus the AWS page. This will require additional dev effort - if significant dev effort is needed will revisit.

lola3736:
Issue re-opened to perform sanity check and review to align Figma and User.  @pixellady FYI, we will further discuss in our team meeting.

pixellady:
Hi @lola3736,
I have completed the sanity check on this Issue 403. No issues noted, good to pass on to Design to create Wireframes.

lola3736:
@pixellady please add "Ready for: Product Lead" label

[...]

PR #538, opened 2023-07-18
Separate sign up route into host and coordinator sign up routes

erikguntner:
This PR is a part of the ongoing refactoring of the sign up experience and includes:

- Creating api spec and separate routes for host and coordinator at `signup/host` and `signup/coordinator`
- Removing the original sign up route
- Updating the current sign up page to make the request to `signup/host` to keep the app functional


api/openapi_server/controllers/auth_controller.py
```diff
@@ -123,6 +123,44 @@ def signup():  # noqa: E501
...
+def signUpCoordinator():  # noqa: E501
+    """Signup a new Coordinator
+    """
+    if connexion.request.is_json:
+        body = connexion.request.get_json()
+
+    secret_hash = get_secret_hash(body['email'])
+
+    # Signup user
+    with Session(db_engine) as session:
```
ju1es, line 135: why do we store a coordinator and not a host in postgres?

api/openapi_server/controllers/auth_controller.py
```diff
@@ -123,6 +123,44 @@ def signup():  # noqa: E501
...
+        user = db.User(email=body['email'])
+        session.add(user)
```
ju1es, line 137: why do we store a coordinator and not a host's email in postgres?

ju1es:
ah is this in preparation for the coordinator invite issue? curious about what the refactoring part is.

api/openapi_server/controllers/auth_controller.py
erikguntner: We save the host as well. Not sure why Github decided not to highlight it in the changes but you can see it if you view the whole file

https://github.com/hackforla/HomeUniteUs/blob/2d1d17d37764cc0ecc231e01b4f2e531252853a3/api/openapi_server/controllers/auth_controller.py

erikguntner:
@ju1es The refactoring relates to these two issues:

https://github.com/hackforla/HomeUniteUs/issues/404
https://github.com/hackforla/HomeUniteUs/issues/403

I think two meetings ago we had a discussion and decided it was best to have separate sign up pages and endpoints after the account selection screen instead of using the same form and endpoint for host and coordinator. This is just the backend portion of that

api/openapi_server/_spec/openapi.yaml
```diff
@@ -420,30 +453,21 @@ paths:
...
           application/json:
             schema:
               type: object
               properties:
-                email:
-                  type: string
-                password:
-                  type: string
-                session:
+                username:
```
mira-kine, line 467: Are we having them create a username only in sign up, or require it for sign in as well?

mira-kine (approved):
Looks good, thanks Erik! I just added a comment about what we might be using "username" in the future about sign up/sign in flow, but other than that looks good. I will work on the front end portion in the upcoming week!

api/openapi_server/_spec/openapi.yaml
erikguntner: @mira-kine thanks for pointing this out. It should be email instead of username. I'll switch it

PR #541, opened 2023-07-19
New password

rochun:
This PR is for the new password page:
- Session_id and user_id are consumed from the URL
- API endpoints are added for the new password request
- Question: Is the password validation component also required for this page?

### DIFF
--- a/app/src/views/NewPassword.tsx
+++ b/app/src/views/NewPassword.tsx
@@ -17,9 +17,11 @@ import {FormContainer} from '../components/authentication';
 
 export const NewPassword = () => {
   const [errorMessage, setErrorMessage] = React.useState('');
+
   const [searchParams] = useSearchParams();
   const session_id = searchParams.get('session_id');
   const user_id = searchParams.get('user_id');
+
   const navigate = useNavigate();
   const [newPassword] = useNewPasswordMutation();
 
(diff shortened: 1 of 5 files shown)


paulespinosa:
@rochun can you link to the issue this is related to please? Thank you.

rochun:
> Reference in
got it

rochun:
> @rochun can you link to the issue this is related to please? Thank you.

got it

app/src/views/NewPassword.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </Typography>
+        <NewPasswordForm
+          onSubmit={handleNewPassword}
+          session_id={session_id}
+          user_id={user_id}
+        />
+        <Divider sx={{width: '100%'}} />
+        <Stack direction="row" alignItems="center" gap={0.5}>
+          <Typography variant="body2">New user?</Typography>
+          <Link underline="always" href="/signup">
```
erikguntner, line 93: Since guests need to be invited, I don't think we need this divider and link to create a new account on this page

rochun: I added it just in case the user was routed to the page without the link, but that makes sense! Thanks.

erikguntner (approved):

ju1es:
awesome work. couple of nits: screenshots? tests?

PR #543, opened 2023-07-20
feat(api): Add endpoint to allow user to resend confirmation email

### LINKED ISSUE #542: Resend confirmation email
### Overview
Users need a way to request resending the confirmation email after signing up.

### Action Items
Create an endpoint that takes the user's email as a parameter and resends the confirmation email.

### Resources/Instructions
Boto3 docs for resending confirmation code:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp/client/resend_confirmation_code.html

Link to the Figma frame:
https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/HUU-Everything-Figma-Nov-2022?type=design&node-id=6764-95252&mode=design&t=qXPmdmA4atJvkami-4

paulespinosa:
A new API endpoint has been added to allow a user to resend a registration confirmation to a user with the provided email.

Limitations:
For the MVP, this commit causes the API to rely on the Cognito service to rate limit the requests rather than keep track of the requests itself.

closes #542


api/openapi_server/controllers/auth_controller.py
```diff
@@ -214,6 +215,43 @@ def signin():
...
+        response = userClient.resend_confirmation_code(
+            ClientId=COGNITO_CLIENT_ID,
+            SecretHash=secret_hash,
+            Username=email,
+        )
+        return response
+    except botocore.exceptions.ClientError as error:
+        match error.response['Error']['Code']:
+            case 'UserNotFoundException':
+                msg = "User not found. Confirmation not sent."
```
erikguntner, line 242: I really like the more specific breakdown of possible errors. Might be worth doing this with the other routes

api/openapi_server/openapi/paths/auth/authResendConfirmationCode.yaml
```diff
@@ -0,0 +1,29 @@
...
+          schema:
+            $ref: '../../openapi.yaml#/components/schemas/ApiResponse'
+    '400':
+      description: The email parameter was not sent or the user with the given email was not found.
+    '429':
+      description: Too many requests to resend the registration confirmation code were made to this user.
+  tags:
+    - auth
+  x-openapi-router-controller: openapi_server.controllers.auth_controller
+  security:
```
erikguntner, line 28: I don't know if we want to require an auth token for this route since users may not be signed in when making the request. For example, post-sign up, but before signing in. I think the way Cognito is currently set up, a user without a verified email won't be able to sign in

erikguntner:
@paulespinosa One thing to note, I checked out the branch to test the endpoint and I ran into an error with the match statements. Apparently, this feature is in versions Python 3.10+, so we might want to change the required version in the docs since it's currently at 3.5.2+

paulespinosa:
Thank you for checking. Yes, we should update the required version.

paulespinosa:
I see. I had thought it was the invitee who would make the request.

api/openapi_server/openapi/paths/auth/authResendConfirmationCode.yaml
paulespinosa: The auth token requirement has been removed.

api/openapi_server/openapi/openapi.yaml
```diff
@@ -16,11 +16,13 @@ paths:
...
     $ref: "./paths/host.yaml"
   /auth/signup/host:
-    $ref: "./paths/auth/authSignupHost.yaml"
+    $ref: "./paths/auth/authSignUpHost.yaml"
   /auth/signup/coordinator:
-    $ref: "./paths/auth/authSignupCoordinator.yaml"
+    $ref: "./paths/auth/authSignUpCoordinator.yaml"
   /auth/signin:
     $ref: "./paths/auth/authSignin.yaml"
+  /auth/resend_confirmation_code:
```
ju1es, line 24: is this suppose to be snake or camel case

ju1es:
preemptively approving but any tests for the controller?

ju1es (approved):

api/openapi_server/openapi/openapi.yaml
paulespinosa: Hi @ju1es! Thank you for taking a look. This is a good and important question. 

### Regarding API endpoint (OpenAPI paths):
I checked again. I see there are both ways being used in this file (e.g. `/serviceProviders`, `/auth/initialInvite`, and `/auth/forgot_password`, `/auth/forgot_password/confirm`). 

* When I searched for guidance in the OpenAPI documentation, I didn't see any explicit convention written, but I did see they used camelCase (this is called mixed case in Python PEP8) in their examples. 
* I then checked https://restfulapi.net/resource-naming/ which only says to be consistent in whatever convention is used and provides "hints" to use hyphen (-) to separate words, avoid underscores, and use lowercase letters. 
* Then, I checked the [connexion documentation](https://connexion.readthedocs.io/en/stable/routing.html#endpoint-routing-to-your-python-views) and saw they had an example using snake case (`/hello_world`) (doh!). 
* Finally, I tried searching for preferences from the React side but couldn't find anything.

Do you, @erikguntner, @Joshua-Douglas  know if there is a preference on the front end side regarding the API endpoint naming convention?

### Regarding OpenAPI operationId: 
I tend to use snake case to follow [PEP8 function and variable name convention](https://peps.python.org/pep-0008/#function-and-variable-names) because it's implemented in Python and the `auth_controller.py` is dominantly snake case.

paulespinosa:
> preemptively approving but any tests for the controller?

I was thinking about this too. Right now, this is a wrapper around the boto3 client, and I was hesitant to mock it. I haven't read all of boto3 documentation yet. So, I'd like some help to write the tests.

api/openapi_server/openapi/openapi.yaml
erikguntner: @paulespinosa Yeah I was using snake case since that's what I saw in the Flask and connexion docs, but I defer to you all with more Python experience for what you wanna do moving forward. Camel case is the standard in Javascript land and always appreciated when getting data back from a db, but the endpoints don't necessarily need to be written in it. Hyphens look fairly clean tho

erikguntner (approved):

api/openapi_server/openapi/openapi.yaml
ju1es: well, at least we have options. thanks for looking into it so thoroughly @paulespinosa!

+1 hyphens for me as well. i don't have a strong preference except that we adhere to one convention for consistency.

ju1es:
> > preemptively approving but any tests for the controller?
> 
> I was thinking about this too. Right now, this is a wrapper around the boto3 client, and I was hesitant to mock it. I haven't read all of boto3 documentation yet. So, I'd like some help to write the tests.

yeah I wasn't sure how to test the controllers which is why I pulled some of the logic out into the 'repositories' class. I think unit tests would be nice. thinking a little more about it, I'm leaning as long as it's been manually tested we can roll with it for now just so we can get things up and running.

Issue #465, opened 2023-02-22
Deploy app to EC2


erikguntner:
### Overview
Test out building and deploying client and server to EC2 for different environments (e.g. dev, prod, test, etc.). Possibly setup Github action to do this in PRs.

### Action Items
**For our existing EC2 instance as of date (2023 July):**

Focus is on dev.homeunite.us, the development/test environment

- [x] Install system-wide python 3.10 for use by the API
- [x] Make application available to a WSGI server
- [x] Manual upload of `.env` containing API configuration until we have another configuration system setup
- [x] Create directory tree: `/opt/homeuniteus-previous-builds/previous` and `/opt/homeuniteus-previous-builds/archives`.
- [x] Create GitHub actions to install the latest version of `main` and run it on EC2. Roughly:
       1. Build and package API as a "wheel" (a binary distribution of a Python package that can be installed directly into an environment)
       2. Upload the package to the EC2 instance
       3. Stop the homeunitus service
       4. Move `/opt/homeuniteus/*` to `/opt/homeuniteus-previous-builds/previous` (the `.env` file remains)
       5. Change directory to `/opt/homeuniteus`
       6. Create a python 3.10 environment in `/opt/homeuniteus`: `python -m venv homeuniteusenv`
       7. Activate the python environment `source homeuniteusenv/bin/activate`
       8.  Install the new API version into the environment: `pip install ~/github-deploy/api/openapi_server-*-.whl`.
       9. Startup the homeuniteus service. If fail, move `/opt/homeuniteus-previous-builds/previous` to `/opt/homeuniteus`, start service, exit 1.
       10. Archive `/opt/homeuniteus-previous-builds/previous` directory into `/opt/homeuniteus-previous-builds/archives/homeuniteus-<date>.tar.gz`  and delete contents of  `/opt/homeuniteus-previous-builds/previous` and delete  older than 3 (arbitrary choice) archived versions from `/opt/homeuniteus-previous-builds/archives/` .

For the frontend app:

- [x] Upload frontend build to the EC2 instance
- [x] Make a backup of the previous files
- [x] move the new frontend build into the directory served by nginx

paulespinosa:
The server is now running Ubuntu 22.04 by upgrading to 20.04 and then to 22.04.

> 5.19.0-1028-aws # 29~22.04.1-Ubuntu SMP Tue Jun 20 19:12:11 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

[...]

PR #544, opened 2023-07-24
refactor(api): Prepare for EC2 deployment

paulespinosa:
The reason for these changes is to prepare the code with the ability to build a
versioned package that can be deployed to a EC2 instance. The build will be done
by CI and the package uploaded to the EC2 instance.

- Updated the minimum Python version to 3.10.
- Made the application available to be run by a WSGI server
- Migrated to `pyproject.toml` declarative project configuration which removes the need for setup.py
- Auto-generated `requirements.txt` and `requirements-dev.txt` based on the core dependencies declared in `pyproject.toml`. These requirements files were created using `pip-compile` from the `pip-tools` project. Instructions on how to update are included in `pyproject.toml`.
- tox.ini updated for minimum version 4.6.4.
- Updated documentation in README files and elsewhere.
- Added a GitHub worfklow to deploy the front-end app and API to the EC2 instance for dev.homeunite.us

Related #465

PR #546, opened 2023-07-25
Create build-deploy-ec2.yml

paulespinosa:
This initial GitHub Workflow, as a manual action, builds and uploads the Home Unite Us front-end application and back-end API web service to the EC2 instance currently used for the Home Unite Us project. 

In this commit, the workflow does not cause the uploaded packages to run. That will come after testing and refining this commit.

related #465

### DIFF
--- a/.github/workflows/build-deploy-ec2.yml
+++ b/.github/workflows/build-deploy-ec2.yml
@@ -100,6 +100,7 @@ jobs:
         with:
           host: homeunite.us
           username: ubuntu
+          key: ${{ secrets.HUU_EC2_SSH_KEY }}
           script_stop: true
           # the following script was manually uploaded
           # this will take at least 2 seconds to complete
--- a/.github/workflows/build-deploy-ec2.yml
+++ b/.github/workflows/build-deploy-ec2.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Install SSH key
         uses: shimataro/ssh-key-action@v2
         with:
-          key: ${{ secrets.SSH_KEY }}
+          key: ${{ secrets.HUU_EC2_SSH_KEY }}
           known_hosts: ${{ secrets.KNOWN_HOSTS }}
       - name: rsync over SSH the App to EC2
         run: |
--- a/.github/workflows/build-deploy-ec2.yml
+++ b/.github/workflows/build-deploy-ec2.yml
@@ -114,7 +114,8 @@ jobs:
             rm ~/github-deploy/api/*
             sudo systemctl restart homeuniteus.service
             cd ~/github-deploy/app
-            sudo mv * /var/www/dev.homeunite.us/html/
+            sudo rsync -a * /var/www/dev.homeunite.us/html/
+            rm -r *
             
 
             
--- a/.github/workflows/build-deploy-ec2.yml
+++ b/.github/workflows/build-deploy-ec2.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Install SSH key
         uses: shimataro/ssh-key-action@v2
         with:
-          key: ${{ secrets.SSH_KEY }}
+          key: ${{ secrets.HUU_EC2_SSH_KEY }}
           known_hosts: ${{ secrets.KNOWN_HOSTS }}
       - name: rsync over SSH the API to EC2
         run: |
--- a/.github/workflows/build-deploy-ec2.yml
+++ b/.github/workflows/build-deploy-ec2.yml
@@ -1,0 +1,87 @@
+name: Build and Deploy to EC2
+#on: workflow_dispatch
+on: pull_request
+jobs:
+  build-deploy-api:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+        working-directory: ./api
+    strategy:
+      matrix:
+        python-version: ["3.10"]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install flake8 tox
+          pip install -r requirements.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/build-deploy-ec2.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Python dependencies
+        working-directory: ./api
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install flake8 tox
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+          if [ -f test-requirements.txt ]; then pip install -r test-requirements.txt; fi
```
Joshua-Douglas, line 22: Aren't `requirements.txt` and `test-requirements.txt` a requirement to properly build and test the application? It might be better to fail if these files aren't found. I'd suggest removing the `-f` check and just do:

```
    pip install -r requirements.txt
    pip install -r test-requirements.txt
```

.github/workflows/build-deploy-ec2.yml
```diff
@@ -0,0 +1,65 @@
...
+        working-directory: ./api
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+      - name: Test with tox
+        working-directory: ./api
+        run: |
+          tox
```
Joshua-Douglas, line 33: Is this the successful build log? `https://github.com/hackforla/HomeUniteUs/actions/runs/5651696696/job/15310200967`

If so, I can't seem to find the `tox` step. Am I overlooking it? I'm mostly wondering because the `api` test cases were failing on my machine. I'd expect them to fail here to, unless I'm missing something

Joshua-Douglas (commented):
Hey @paulespinosa,

Looks great! Have you been deploying manually up to this point? If so, I'm sure this will make your life a lot easier. 

I just have a few minor comments and questions.

.github/workflows/build-deploy-ec2.yml
Joshua-Douglas: 🤦‍♂️That link I posted was the PR build link haha. 

Have you tried out the workflow and confirmed that it worked? If so, can I take a look at the deployment log?

Thanks!

.github/workflows/build-deploy-ec2.yml
paulespinosa: That's better! Thank you.

.github/workflows/build-deploy-ec2.yml
paulespinosa: :blush:  Actually, I'm currently figuring out how to test it without merging to the main branch.

paulespinosa:
> Hey @paulespinosa,
> 
> Looks great! Have you been deploying manually up to this point? If so, I'm sure this will make your life a lot easier.
> 
> I just have a few minor comments and questions.

As far as I can tell, we haven't been deploying code for quite a while. I think Tyler was manually doing the deployments in the past.

.github/workflows/build-deploy-ec2.yml
Joshua-Douglas: Have you tried out this workflow yet? It might be worth sending off a build before approving, to ensure the SSH access is working. 

Do we have a development EC2 instance? If we did, then we wouldn't have to worry about breaking the current site

Joshua-Douglas:
> As far as I can tell, we haven't been deploying code for quite a while. I think Tyler was manually doing the deployments in the past.

Ah, that is great to know. Thanks. Now I've got a bunch of questions about the version of our deployment and our deployment strategy. I'll ask at today's meeting! 

Either way, thanks for getting this started!

.github/workflows/build-deploy-ec2.yml
paulespinosa: @Joshua-Douglas Yes, I've got it in a testable state and have been modifying it. You can see all of the runs in the Actions tab.

At this time, we have just one EC2 instance. It hosts the static landing page for https://homeunite.us/ and acts as the dev/test/demo host for https://dev.homeunite.us. As of this writing, it's hosting an old version of the frontend and the backend is down.

paulespinosa:
> > current working directory "/home/runner/work/HomeUniteUs/HomeUniteUs"
> 
> @paulespinosa you may need to set the working directory to `app` for the end to end tests in the build-deploy-app workflow.

@erikguntner Thank you for catching that. I added it back in and it looks like its working again.

ju1es:
sweet. seeing in the [build](https://github.com/hackforla/HomeUniteUs/actions/runs/5663719483/job/15345905692?pr=546) that it's failing on the service provider controller tests. I think Joshua has a [PR](https://github.com/hackforla/HomeUniteUs/pull/547) that should fix. 

I forget if we're working towards ci/cd but either way, can we update the readme with instructions for deployment.

paulespinosa:
@tylerthome  For this workflow, I used the shimataro ssh key action with the following configuration:

```
      - name: Install SSH key
        uses: shimataro/ssh-key-action@v2
        with:
          key: ${{ secrets.SSH_KEY }}
          known_hosts: ${{ secrets.KNOWN_HOSTS }}
```
When the workflow runs, it appears that the shimataro ssh key action isn't picking up the `SSH_KEY` or `KNOWN_HOSTS` secrets. The error is "Error: Error: Input required and not supplied: key". Am I using the right variable names?

Also, it looks like we'll need to add the .env variables to GitHub secrets because the workflow runs tox as a step. When the tests runs the `auth_controller.py` module is loaded. This module has a top-level instantiation of a boto3 client that checks for the AWS credentials during initialization. Since the environment doesn't have the AWS credentials, the boto3 client instantiation fails causing the tests to fail.

paulespinosa:
The workflow is being added to the PR #544 because the changes are related. This PR is being closed.

Issue #539, opened 2023-07-18
API: Remove dependency on swagger-cli bundle


paulespinosa:
### Overview
The API currently needs to bundle the OpenAPI specification YAML files located in the API source code under `openapi_server/openapi` into a single file using the `swagger-cli bundle` tool because the `connexion` framework does not support the specifications spread across files. This issue is to remove this requirement so that the OpenAPI specification files in `openapi_server/openapi` can be used directly by the API code.

### Action Items
Research getting `connexion` to use the `openapi_server/openapi` spec files. Take a look at the comment on the connexion issue [Breaking up large swagger.yaml files
](https://github.com/spec-first/connexion/issues/254#issuecomment-1203427194) for ideas.

### Resources/Instructions
https://github.com/spec-first/connexion/issues/254 - a connexion issue about this problem
https://python-jsonschema.readthedocs.io/en/stable/referencing/ - the underlying library that parses the specs
https://referencing.readthedocs.io/en/stable/ - a referencing library used by jsonschema might have the ability to piece together separate files

erikguntner:
Looks like [prance](https://github.com/RonnyPfannschmidt/prance) could potentially be what we are looking for.
https://github.com/spec-first/connexion/issues/254#issuecomment-497194240

paulespinosa:
> Looks like [prance](https://github.com/RonnyPfannschmidt/prance) could potentially be what we are looking for. [spec-first/connexion#254 (comment)](https://github.com/spec-first/connexion/issues/254#issuecomment-497194240)

This looks very promising.

erikguntner:
Looks like this is fixed in Connexion 3, which is currently in Alpha.

https://github.com/spec-first/connexion/pull/1648

erikguntner:
It worked! I'll try to open a PR tonight. I just need to replace where _spec was being referenced and replace it.

paulespinosa:
> Looks like this is fixed in Connexion 3, which is currently in Alpha.
> 
> [spec-first/connexion#1648](https://github.com/spec-first/connexion/pull/1648)

It's good to see they are addressing this. Too bad it's still in Alpha at the moment, but can be something to look forward to.

PR #547, opened 2023-07-25
Fix API Test Case Failures

Joshua-Douglas:
The API test cases are failing with this error: `jsonschema.exceptions.RefResolutionError: unknown url type: './paths/users.yaml'`.

The root cause of this failure was in `BastTestCase.create_app`. There is a known issue related to the application specification in 
#539. The specification file cannot contain references to other files, so our flask app should use the `openapi_server/_spec/openapi.yaml` file. The test app, however, was using the `openapi_server/openapi/openapi.yaml` file instead.

Unfortunately, when I made this fix I found several other failing test cases and some other sources of testing errors. This PR contains the minimum fixes needed to get the test cases passing again. I refrained from making any design changes at this point.

Here are a summary of the fixes:
* Fix an `AttributeError` in `service_provider_controller.py`'s `create_service_provider()` method. The test cases were failing with a status code of `500` since an unhandled exception was being thrown here
* Fix `Instance not bound to session` err that was being thrown while creating a service provider. Test cases were failing with a status code of `500` here too.
  * The database row id was being read after calling `session.close()`. This is not a problem if the session has been flushed, but by default SQLAlchemy will lazily load the row, so attempting to read after closing the session can cause an error. To fix I just read the row id before closing.
* Fix return syntax error in `get_service_provider_by_id` and `update_service_provider`. These methods were not returning the correct status code if the provider search failed
* Add an error status code to `delete_service_provider` if the provider deletion fails. Our test cases were passing, even though no deletion was occuring, since we were always passing back a successful status code
* Update the controller test cases to use a temporary testing database. Previously the test cases were operating on the host's local database. This means that the tests would only pass if the host happened to have a database with a sufficient number of entries. The new test setup has the added benefit of making the controller test cases independent of each other.
  * Note: Our Flask app interface does not allow the ability to specify a custom database, so I had to hack it a bit. Ideally we'd redesign our interface with testing in mind - but I wanted to limit the scope of this issue 
* Fix syntax error in `HousingProviderRepository.__init__`. Default args are only evaluated when the method definition is executed, so move the call to `get_engine` into the method body. 
* remove `test_match_filtering.py`. This file was empty and unused

### DIFF
--- a/api/openapi_server/controllers/service_provider_controller.py
+++ b/api/openapi_server/controllers/service_provider_controller.py
@@ -42,7 +42,7 @@ def delete_service_provider(provider_id):  # noqa: E501
     """
     if housing_provider_repository.delete_service_provider(provider_id):
         return f"Service Provider with id {provider_id} successfully deleted", 200
-    return f"Service provider {provider_id} not found", '404'
+    return f"Service provider {provider_id} not found", 404
 
 def get_service_provider_by_id(provider_id):  # noqa: E501
     """Get details about a housing program service provider from an ID
--- a/api/openapi_server/controllers/service_provider_controller.py
+++ b/api/openapi_server/controllers/service_provider_controller.py
@@ -26,7 +26,7 @@ def create_service_provider():  # noqa: E501
     except ValueError:
         return traceback.format_exc(ValueError), 400
     
-    withId = housing_provider_repository.createServiceProvider(provider)
+    withId = housing_provider_repository.create_service_provider(provider)
     return withId, 201
 
 
--- a/api/openapi_server/controllers/service_provider_controller.py
+++ b/api/openapi_server/controllers/service_provider_controller.py
@@ -54,7 +54,7 @@ def get_service_provider_by_id(provider_id):  # noqa: E501
     :rtype: ServiceProviderWithId
     """
     provider = housing_provider_repository.get_service_provider_by_id(provider_id)
-    return provider, 200 if provider != None else "Not Found", 404
+    return (provider, 200) if provider is not None else ("Not Found", 404)
 
 def get_service_providers():  # noqa: E501
     """Get a list of housing program service providers.
@@ -89,4 +89,4 @@ def update_service_provider(provider_id):  # noqa: E501
         return traceback.format_exc(ValueError), 400
     
     updated = housing_provider_repository.update_service_provider(provider, provider_id)
-    return updated, 200 if updated != None else "Not Found", 404
+    return (updated, 200) if updated is not None else ("Not Found", 404)
--- a/api/openapi_server/controllers/service_provider_controller.py
+++ b/api/openapi_server/controllers/service_provider_controller.py
@@ -40,8 +40,9 @@ def delete_service_provider(provider_id):  # noqa: E501
 
     :rtype: None
     """
-    housing_provider_repository.delete_service_provider(provider_id)
-    return f"Service Provider with id {provider_id} successfully deleted", 200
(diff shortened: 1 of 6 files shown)


Joshua-Douglas:
Just a heads up - the PR build failed the first time I sent it off. I think it is a sporadic, unrelated failure. 

Here was the error in the `npm run test` stage. The build passed when I restarted it.

Thanks!

```
 ❯ src/views/__tests__/ForgotPasswordCode.test.tsx:55:5
     53| 
     54|     expect(screen.queryByRole('alert')).not.toBeInTheDocument();
     55|     waitFor(() => expect(screen.getByRole('form')).toBeInTheDocument()…
       |     ^
     56|   });
     57| 
 ❯ node_modules/@vitest/runner/dist/index.js:160:13
 ❯ node_modules/@vitest/runner/dist/index.js:41:26
 ❯ runTest node_modules/@vitest/runner/dist/index.js:491:17
 ❯ runSuite node_modules/@vitest/runner/dist/index.js:594:15
 ❯ runSuite node_modules/@vitest/runner/dist/index.js:594:15
 ❯ runFiles node_modules/@vitest/runner/dist/index.js:645:5
 ❯ startTests node_modules/@vitest/runner/dist/index.js:655:3
 ❯ node_modules/vitest/dist/entry.js:276:7

This error originated in "src/views/__tests__/ForgotPasswordCode.test.tsx" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "display an error message". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- This was the last recorded test before the error was thrown, if error originated after test finished its execution.
Error: Process completed with exit code 1.
```

paulespinosa (approved):
The changes look good. Thank you.

paulespinosa:
@Joshua-Douglas I also noticed that the npm tests fail sporadically.

erikguntner:
@Joshua-Douglas @paulespinosa I'll take a look at the tests. It seems to be that same one that fails randomly.

erikguntner (approved):

ju1es:
sweet, thank you for fixing these! awesome test scaffolding. I'm not used to writing tests for endpoints that aren't part of a testable class which is why I pulled a lot of the logic out into the repository class. 

preemptively approving but just one question below,

> Fix syntax error in HousingProviderRepository.__init__. Default args are only evaluated when the method definition is executed, so move the call to get_engine into the method body.
i'm not a pythonista so just want to clarify, is the function application in the body also evaluated during function execution?

ju1es (approved):

Joshua-Douglas:
> i'm not a pythonista so just want to clarify, is the function application in the body also evaluated during function execution?

Hey, happy to clarify! The body is  executed each time the function is called, but the default arg assignment only happens once. 

The general rule for default args is that they should be set to an immutable type, since otherwise the default arg will be a reference to a mutable object that may change each time the function is called. This leads to unexpected behavior, so 'None' is typically used in place of a default mutable type.

Here are some examples the demonstrate the weird behavior.

I hope these help! If not I'm sure I can find an article to explain it better.
'''
def append_print(el, lst=list()):
    lst.append(el)
    print(lst)
    
append_print(1) # print [1]
append_print(2) # print [1,2]?!
append_print(3, list()) # print [3]

class DBLayer:
    engine = None
    
def print_engine(eng=DBLayer.engine):
    print(eng)
    
print_engine() # None, as expected 
DBLayer.engine = "Fancy engine"
print_engine() # Still None!!
print_engine(DBLayer.engine) # Prints "Fancy engine"
'''

PR #548, opened 2023-07-25
Sign in loading states

erikguntner:
Currently there aren't any loading states when a user was signing in, so this PR adds a spinner and disables the submit button when making a request either with email and password or with Google.


ju1es:
screenshots?

ju1es (approved):

Issue #549, opened 2023-07-25
Dropdown menu in nav bar stays open and loses anchor point when logging out


erikguntner:
### Overview
If you log out through the avatar dropdown menu, the menu will lose it's anchor point and be automatically open the next time you log in. It should start off closed when starting a new session.
<img width="1287" alt="Screenshot 2023-07-25 at 10 54 39 AM" src="https://github.com/hackforla/HomeUniteUs/assets/27253583/4beadf1d-8fe5-4c60-992a-02cf874a2181">

PR #550, opened 2023-07-25
Fix dropdown menu bug

erikguntner:
Fixes [549](https://github.com/hackforla/HomeUniteUs/issues/549)

The issue came from setting the anchor element for the menu as the avatar, but then removing the element when the user logs out and not clearing the state. One approach would have been to reset the state when logging out, but I felt it made sense to break this into it's own component so it can handle it's own state.


ju1es:
I'm not sure what the state of testing is on the frontend but do we have a test for this? in case it comes up again we can catch it.

ju1es (approved):

PR #554, opened 2023-07-26
Remove dependency on swagger cli bundle

erikguntner:
Fixes [539](https://github.com/hackforla/HomeUniteUs/issues/539)

This PR uses the library [Prance](https://github.com/RonnyPfannschmidt/prance) to parse and resolve paths used in the OpenAPI spec so we no longer have to generate one. 

The solution was suggested in [this Github issue](https://github.com/spec-first/connexion/issues/254#issuecomment-497194240)

I also looked into incorporating [OpenAlchemy](https://openapi-sqlalchemy.readthedocs.io/en/latest/index.html) which auto-generates SQLAlchemy models from the spec, but I ran into an issue with conflicting packages and decided to save that for another PR.


api/openapi_server/test/__init__.py
```diff
@@ -33,9 +34,11 @@ def create_app(self):
         database.DataAccessLayer.db_init()
 
         logging.getLogger('connexion.operation').setLevel('ERROR')
-        app = connexion.App(__name__, specification_dir='../_spec/')
+        app = connexion.App(__name__)
         app.app.json_encoder = JSONEncoder
-        app.add_api('openapi.yaml', pythonic_params=True)
+        app.add_api(get_bundled_specs(Path('openapi_server/openapi/openapi.yaml')),
+                arguments={'title': 'Home Unite Us'},
+                pythonic_params=True)
```
Joshua-Douglas, line 41: This code is duplicated from `__main__.py`. Have we thought about add an `App` class to encapsulate this logic? If we did it that way then `__main__.py` and `BaseTestCase` would just need to create an instance of our app class, instead of replicating this kind of complex logic. 

If we go that route I'd probably make `get_bundled_specs` a "private" method within the app class.

Joshua-Douglas (changes requested):
Hey @erikguntner - Great job!! Solution looks really great to me. 

Now that we no longer need to generate the flattened yaml file, can you delete the `_spec` folder? Thanks!

api/openapi_server/test/__init__.py
Joshua-Douglas: You can choose a better name than just `App`, of course

paulespinosa: @Joshua-Douglas @erikguntner I can make that change in the 465-deploy-to-ec2 after this PR goes to main.

erikguntner:
@Joshua-Douglas Deleted! Thanks for catching that

api/openapi_server/test/__init__.py
erikguntner: @paulespinosa Sounds good to me, thanks!

paulespinosa:
Hasn't anyone gotten this error yet:

> prance.util.url.ResolutionError: (2, 'No such file or directory', 'File not found: /Development/HackForLA/HomeUniteUs/api/openapi_server/openapi/paths/auth/authSignupHost.yaml')

The filename specified in openapi.yaml is different than what's in the filesystem (authSignUpHost.yaml). Same for authSignupCoordinator.yaml <=> authSignUpCoordinator.yaml.

paulespinosa (approved):
This a great step forward! The app starts up and responds to requests for me when I run `python -m openapi_server` in the `api` directory.  Thank you! :rocket:

Joshua-Douglas:
> Hasn't anyone gotten this error yet:
> 
> 
> 
> > prance.util.url.ResolutionError: (2, 'No such file or directory', 'File not found: /Development/HackForLA/HomeUniteUs/api/openapi_server/openapi/paths/auth/authSignupHost.yaml')
> 
> 
> 
> The filename specified in openapi.yaml is different than what's in the filesystem (authSignUpHost.yaml). Same for authSignupCoordinator.yaml <=> authSignUpCoordinator.yaml.
> 
> 

Are you still seeing this error? I didn't see it. What command did you run? I can try to replicate

paulespinosa:
> > Hasn't anyone gotten this error yet:
> > > prance.util.url.ResolutionError: (2, 'No such file or directory', 'File not found: /Development/HackForLA/HomeUniteUs/api/openapi_server/openapi/paths/auth/authSignupHost.yaml')
> > 
> > 
> > The filename specified in openapi.yaml is different than what's in the filesystem (authSignUpHost.yaml). Same for authSignupCoordinator.yaml <=> authSignUpCoordinator.yaml.
> 
> Are you still seeing this error? I didn't see it. What command did you run? I can try to replicate

I ran `python -m openapi_server` in the `api/` directory.

Joshua-Douglas:
Hey @erikguntner, @paulespinosa alerted me to the fact that app.run() is designed to only run the development server. It got me wondering - how do we pass the parsed api specification to the production server?

erikguntner:
@Joshua-Douglas Good question. I'm not totally sure. I'd check the Flask docs for a building for production section or something. One option that comes to mind is generating the combined spec for a production build and saying if we're in a production environment use the generated spec? But I don't know if we wanna do that right after we tried to get away from generating one haha

Joshua-Douglas:
> @Joshua-Douglas Good question. I'm not totally sure. I'd check the Flask docs for a building for production section or something. One option that comes to mind is generating the combined spec for a production build and saying if we're in a production environment use the generated spec? But I don't know if we wanna do that right after we tried to get away from generating one haha

Haha agreed. I'll take a look tonight or tomorrow and get back to you! 

This might also be a good distinction to capture in the README. I'm updating it now, so I'll add the different deployment instructions (once I figure it out haha)

api/openapi_server/__main__.py
```diff
@@ -22,15 +26,22 @@
...
-    app = connexion.App(__name__, specification_dir='./_spec/')
+    app = connexion.App(__name__)
 
     # Set configs
     env_config_profile = get_key(find_dotenv(), 'CONFIG_PROFILE')
     app.app.config.from_object(compile_config(env_config_profile))
     app.app.json_encoder = encoder.JSONEncoder
     app.app.secret_key = SECRET_KEY
-    app.add_api('openapi.yaml',
+    app.add_api(get_bundled_specs(Path('openapi_server/openapi/openapi.yaml')),
```
paulespinosa, line 44: @erikguntner @Joshua-Douglas The code below will work in production and dev:

```
    app.add_api(get_bundled_specs(app.get_root_path() / Path('./openapi/openapi.yaml')),
                arguments={'title': 'Home Unite Us'},
                pythonic_params=True)             
```

paulespinosa (changes requested):

paulespinosa (changes requested):
@erikguntner ~In the `openapi/openapi.yaml` file, please uppercase the 'u' in "up" in the ref paths like the following (just these two refs) because on my system file names are case-sensitive. When written as lowercase, the files aren't found.:~

Edit: I pushed a change to update this file.
```
  /auth/signup/host:
    $ref: "./paths/auth/authSignUpHost.yaml"
  /auth/signup/coordinator:
    $ref: "./paths/auth/authSignUpCoordinator.yaml"

```

paulespinosa:
@erikguntner @Joshua-Douglas I pushed a couple of changes to this branch. One change allows the `openapi.yaml` file to be found in a production environment. We don't need to flatten the spec for production or anything else special for production. The other change fixes the reference paths to match the filenames. Thank you.

Joshua-Douglas:
> how do we pass the parsed api specification to the production server?

Thanks @paulespinosa for giving me the answer to this question over slack. 

The development and production server are started using different methods, but both server types using the `app_api()` method to set the api specification - so your changes should work in both contexts!

The distinction between starting a development and production server is explained fully on #544, along with the source code changes needed to properly run our project using `gunicorn`.

With that, everything looks good on my end. Approved!

Joshua-Douglas (approved):

paulespinosa (approved):

Issue #520, opened 2023-06-06
Section 2 - Coordinator - Invite Guest


rpradheap:
Problem Alignment

The Problem

Coordinators should be able to create a user profile and send invitation to guest

Application that are not yet submitted

User Story:

As a coordinator, I want to create a profile for guest and send the invite email

High Level Approach

Coordinator is able to invite guest to submit an application via HUU system

Solution Alignment

Considerations

Coordinators will be using desktop or tablet--> use responsive design

Acceptance criteria
Success is if Coordinator can successfully send a invite to Guest

Key Features

Add Guest Button - Coordinator is able to add the guest email id and send a invite to guest with a predefined email content that states how to get started with the application

Design
https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/HUU-Everything-Figma-Nov-2022?node-id=2353%3A17352&t=ZM7AykeF7WVNoXkD-0

 Review Existing Design
 Updating Design as Needed
Limitations

None

pardhama:
Check with engineering regarding personalization of email text and work on the draft

sanya301:
Check if invites expire, and define a flow in the story for what happens if it does expire

erikguntner:
The temporary passwords in the invite do expire. The default is 7 days and can be adjusted all the way up to 365 days. I'll look into sending new invites.

stevbark:
I can work on this for the Engineering side. One note is that it says this has both front and back end components and I am focused on backend atm.

edela0015:
Quick question: are we giving the coordinator the ability to revoke the invite as well? or resend? thank you.

brycelednar:
Hi @edela0015, 

Is there a user scenario that you are envisioning for revoke invite? If we think it will be an infrequent occurrence and it's not critical, it's ok to leave out in the first version.

For resend, my take is that we keep it simple/low effort in this version and just allow Coordinators to go through the user flow again to invite a user, versus build dedicated re-send invite functionality.

KowDesign:
https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/HUU-Everything-Figma-Nov-2022?type=design&node-id=5872-72669&mode=design&t=6P4UHiA7FqyFeQuC-4

rpradheap:
@KowDesign - Please check my comment in Figma about the Invite Guest Button placement. I think the pages once we click the invite - name and email looks good .

erikguntner:
[...]

PR #558, opened 2023-08-01
Guest invite

rochun:
This invite guest component PR (#520) includes:

- Coordinator API endpoint and file
- Guest Invite button and modal
- Input validation for name and email

Screen Shot 2023-08-01 at 11 53 40 AM

Screen Shot 2023-08-01 at 11 53 49 AM


app/src/components/common/guestInvite.tsx
```diff
@@ -0,0 +1,152 @@
...
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+} from '@mui/material';
+import {createTheme, ThemeProvider} from '@mui/material/styles';
+import ControlPointIcon from '@mui/icons-material/ControlPoint';
+import {useInviteGuestMutation} from '../../services/coordinator';
+
+const theme = createTheme({
```
erikguntner, line 18: We shouldn't have to create a theme and theme provider just for styling this button. You can sx or styled components to make these changes.

app/src/components/common/guestInvite.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <ControlPointIcon color="primary" sx={{marginRight: '0.25em'}} />
+          Invite Guest
+        </Button>
+
+        <form onSubmit={handleSubmit}>
+          <Dialog
+            open={open}
+            onClose={handleClose}
+            fullWidth={true}
+            maxWidth={'xs'}
```
erikguntner, line 88: I think you can remove the brackets here

app/src/components/common/guestInvite.tsx
```diff
@@ -0,0 +1,152 @@
...
+                  id="outlined-name-input"
+                  label="Full Name"
+                  type="FullName"
+                  name="fullName"
+                  value={fullName}
+                  onBlur={handleBlur}
+                  onChange={handleChange}
+                  error={touched.fullName && Boolean(errors.fullName)}
+                />
+                {touched.fullName && errors.fullName && (
```
erikguntner, line 109: There should be a helper text property you can use on TextFields

app/src/components/common/guestInvite.tsx
```diff
@@ -0,0 +1,152 @@
...
+                  name="email"
+                  sx={{
+                    marginTop: '1em',
+                  }}
+                  value={email}
+                  onChange={handleChange}
+                  onBlur={handleBlur}
+                  error={touched.email && Boolean(errors.email)}
+                />
+                {touched.email && errors.email && (
```
erikguntner, line 125: Same as above about the helper text

erikguntner:
@rochun looks good so far! Just a couple of changes

app/src/components/common/guestInvite.tsx
erikguntner: To add a little more context. Looks like this was done to add the primary color from Figma to the buttons? We still haven't gotten a definite answer on whether this is the color we are going with so don't worry about if for now. When we get the okay, we'll update the color in the main theme and the changes should be applied. It would probably be good to check with design over the next week.

rochun:
> @rochun looks good so far! Just a couple of changes

@erikguntner thanks for the feedback! I'll make the changes. 🦾

rochun:
@erikguntner would you suggest I modularize the input validation component like we did the others?

PR #573, opened 2023-08-19
Dashboard outer layout

erikguntner:
This PR adds the wrapping layout for the guest dashboard, including the header, side nav, and main section for displaying the current page.

https://github.com/hackforla/HomeUniteUs/assets/27253583/bb38a15e-78a7-45c7-9998-648a7cd69f60

Potential future changes:
- Currently, the navigation links for the side nav are hard coded in, but it could be passed in as a prop depending on if the host dashboard's links are different.


app/src/components/layouts/DashboardLayout.tsx
```diff
@@ -0,0 +1,106 @@
...
+  padding: `${theme.spacing(5)} ${theme.spacing(2)} 0 ${theme.spacing(2)}`,
+  gap: theme.spacing(3),
+}));
+
+const StyledListItemButton = styled(ListItemButton, {
+  shouldForwardProp: prop => prop !== 'isSelected',
+})<ListItemProps>(({isSelected, theme}) => ({
+  gap: 1,
+  borderRadius: theme.shape.borderRadius,
+  color: isSelected ? 'black' : 'white',
```
ju1es, line 82: and here? any attribute that uses color

ju1es:
I pulled your PR, trying to play with it locally. Do you have instructions for logging in/accessing the `/guest` route? Also, does this PR incorporate styling for mobile?

erikguntner:
@jules Here is a test user that I added to Cognito that you can use to log in:

email: testuser@email.com
password: Test1234!

Just make sure to add the email to your user table first for it to work.

Right now I'm just typing `/guest` into the url. In the future when someone logs in as a guest they will be redirected here, but since we don't have roles set up yet, do you think I should add a link in the header to make it easier to navigate to?

erikguntner:
@ju1es This doesn't include mobile right now, but I see that there are now designs for it so I'll make these updates

Joshua-Douglas:
I tried it out and it works great! 

We may need a more responsive side panel to properly support mobile, instead of using `const drawerWidth = 209;`. I tried to login using my phone but the 'sign in' and 'sign up' buttons don't work from my iPhone for some reason. 

Other than that everything looks good to me!

JpadillaCoding (commented):
@erikguntner Everything looks good on the outer layout. Am I correct to assume that the inner layout will be done on a separate branch?

erikguntner:
@JpadillaCoding that's correct. This is just the outer layout portion

JpadillaCoding:
@erikguntner Great. I haven't found anything to improve that hasn't already been pointed out so I'll give my review approval.

JpadillaCoding (approved):

mira-kine (approved):
Looks good! As for the comments about the colors, we only have the dark/light themes and then the themed buttons which use black outlines. If the design requires just black/white for accessibility purposes, then I think we can call it a one off?

erikguntner:
I made a few more updates to this PR:

- I made sure colors came from the theme provider.
- Created an authenticated header separate from the main header
- Moved the mobile menu to the dashboard layout component and rendered the links from the passed in nav links
- The dashboard layout now accepts the nav items as a prop so they can be passed in and different between user types

Issue #569, opened 2023-08-16
Task component


erikguntner:
### Overview
A task that displays inside the task accordion when expanded.

<img width="491" alt="Screenshot 2023-08-16 at 2 28 58 PM" src="https://github.com/hackforla/HomeUniteUs/assets/27253583/c049f6b4-5c76-4264-b885-ba3630cc3fe2">

### Action Items
- Create a reusable task component that accepts a set of props that allows for it's varied state and contents

Issue #568, opened 2023-08-16
Task accordion component


erikguntner:
### Overview
Create an accordion component that can expand to contain a list of tasks. The border on the image wraps the whole list of accordion task components, but this should just focus on the single component itself.

<img width="499" alt="Screenshot 2023-08-16 at 2 28 44 PM" src="https://github.com/hackforla/HomeUniteUs/assets/27253583/8d6421db-fff0-41dc-bca8-b40be23ba351">

### Action Items

- Create a reusable accordion component
- Define a set of props that can be passed to make this reusable
- For rendering the list I think this could either be done passing a list of tasks or a children prop

### Resources/Instructions
MUI accordion component: https://mui.com/material-ui/react-accordion/

Issue #391, opened 2022-04-06
Design and implement initial state machine database model for all business entities


tylerthome:
### Overview
Define the base schema(s) to be used for stateful entities in the HUU logical business model

### Action Items
- [ ] Create a state machine or statechart for the `Case` model

### Resources/Instructions
* Example basic statechart information for `Case`:
  * states:
    * awaiting_coordinator_approval
    * awaiting_host_acceptance
    * awaiting_initial_meeting
    * ineligible
  * events
    * guest_submit_interest
    * guest_withdraw_interest
    * coordinator_approve
    * coordinator_reject
    * coordinator_create_case
    * host_accept
    * host_reject
* Reference material:
  * For design, consider the `Trip` model in [this article](https://eng.uber.com/sessionizing-data/)
  * For implementation, consider the "Programming Model" section of [this article](https://eng.uber.com/fulfillment-platform-rearchitecture/)
  * Conceptual and technical
    * https://xstate.js.org/docs/guides/introduction-to-state-machines-and-statecharts
    * https://statecharts.dev/
    * https://en.wikipedia.org/wiki/Finite-state_machine

tylerthome:
An important requirement to bear in mind for this task is that we will need two sets of dynamic state types that will vary by organization:
* Document submittals
* Trainings / orientations

For the initial pass at the state machine or statechart, consider these as aggregate states e.g. `awaiting_all_documents_submitted` and `awaiting_all_trainings_completed` (feel free to change the naming convention, another alternative for these examples could be `host_submitting_documents` and `host_completing_trainings`  or any that clearly represent the state)

tylerthome:
As part of the existing work underway for the basic features, this requirement will be satisfied using the PostgreSQL transaction log directly, or an integration with e.g. [RedHat's Debezium](https://debezium.io/documentation/reference/2.6/connectors/postgresql.html)

tylerthome:
Design has been established and future work specification is underway

lola3736:
@tylerthome please review the complexity label for accuracy.

PR #596, opened 2023-09-14
Task accordion and sub tasks

erikguntner:
https://github.com/hackforla/HomeUniteUs/assets/27253583/a5629fd5-fabe-41b4-bad0-7aa3b2461c39

This PR includes the following changes:

- Adds task accordion components to the My Tasks section on the dashboard homepage
- Task accordion components can be opened by clicking on them, displaying a list of what I've called subtasks.
- Task accordions display a count of how many subtasks have been completed
- Task accordions render different elements based on the status of the task
- Task accordion with the "in progress" task is expanded on page load
- Subtasks display a title, description, and icon that changes based on the current status.
- Subtasks also display either a button or message based on the status

Still to do:

- Components are using dummy data so this will need to be replaced with an API call when the endpoints are in place.

Closes [568](https://github.com/hackforla/HomeUniteUs/issues/568) Closes [569](https://github.com/hackforla/HomeUniteUs/issues/569)

Below are the Typescript types representing the general structure:

```
// The status of a task
export type TaskStatus = 'in-progress' | 'complete' | 'locked';

export interface Task {
  id: number;
  title: string;
  status: TaskStatus;
  subTasks: SubTask[]; // array of subtasks
}

export interface SubTask {
  id: number;
  title: string;
  status: TaskStatus;
  description: string;
  buttonText: string;
  url: string;
}
```
An example of a Task object:
```
  {
    id: 1,
    title: 'Application and Onboarding',
    status: 'in-progress',
    subTasks: [
      {
        id: 1,
        title: 'Application',
        status: 'complete',
        description:
          'Start your guest application to move on to the next step.',
        buttonText: 'Start application',
        url: '/guest-application',
      },
      {
        id: 2,
        title: 'Coordinator Interview',
        status: 'in-progress',
        description: 'Meet with your Coordinator to share more about yourself.',
        buttonText: 'Schedule interview',
        url: '/schedule',
      },
      {
        id: 3,
        title: 'Training Session',
        status: 'locked',
        description:
          'Complete a training session to prepare you for the host home experience.',
        buttonText: 'Schedule training',
        url: '/schedule',
      },
    ],
  },
```

My idea for this structure came partly from the design requirements but also from issue [391](https://github.com/hackforla/HomeUniteUs/issues/391) and how the state machine and events will move the user through this process. The main tasks and subtasks all have IDs that correspond to various states in the state machine and their status would then be updated as different events occur. For example, after a user submits their application and it's approved, an event would fire changing the status of the `submit_application` task to "complete" and simultaneously changing the status of ` interview_with_coordinator` to  "in-progress".

I'd love to get your feedback on the structure and my current approach, as well as what can be done to improve it. Let me know if the properties and names need to be changed, what could be added or removed, or if this is just way off.  I hope it can start a discussion about how we want to structure the data on the backend and what is needed on the frontend. Thanks!

```mermaid
erDiagram
    USER ||--|{ TASK : has
    TASK {
        int id PK
        int userId FK
        string title
        int statusId FK 
    }
    TASK ||--|{ SUB_TASK: contains
    TASK ||--|| STATUS: has
    SUB_TASK ||--|| STATUS: has
    SUB_TASK{
        int id PK
        int taskId FK
        string title
        string description
        int statusId FK
        string buttonText 
        url string
    }
    STATUS {
       int id PK
       string status "locked, inProgress, complete" 
    }
```


app/src/views/GuestApplicationTracker.tsx
```diff
@@ -1,5 +1,97 @@
...
+export interface Task {
+  id: TaskIds;
+  title: string;
+  status: TaskStatus;
+  subTasks: SubTask[];
+}
+
+const tasks: Task[] = [
+  {
+    id: 'application_and_onboarding',
```
paulespinosa, line 36: Will it be ok if the `id` is an integer?

app/src/views/GuestApplicationTracker.tsx
```diff
@@ -1,5 +1,97 @@
...
+    title: 'Application and Onboarding',
+    status: 'in-progress',
+    subTasks: [
+      {
+        id: 'submit_application',
```
paulespinosa, line 41: Same question about `id` being integer here and other places as well.

app/src/views/GuestApplicationTracker.tsx
```diff
@@ -1,5 +1,97 @@
...
+  status: TaskStatus;
+  subTasks: SubTask[];
+}
+
+const tasks: Task[] = [
+  {
+    id: 'application_and_onboarding',
+    title: 'Application and Onboarding',
+    status: 'in-progress',
+    subTasks: [
```
paulespinosa, line 39: Do the subtasks need a link or other data to let the FE know where to send the user to complete the subtask?

app/src/components/dashbaord/DashboardTask.tsx
```diff
@@ -0,0 +1,77 @@
...
+  console.log('do something');
+};
+
+interface TaskInfo {
+  title: string;
+  onClick?: () => void;
+}
+
+const taskInfo: {[key in SubTaskIds]: TaskInfo} = {
+  submit_application: {
```
paulespinosa, line 17: As I understand the set of tasks is fixed for the MVP. Is that correct?

paulespinosa (commented):
It's looking good! I just need to do a little more background research on the feature to know more about the number of tasks, subtasks, etc. to expect. Thank you!

app/src/views/GuestApplicationTracker.tsx
erikguntner: @paulespinosa Yes, if that's something we could keep in the db that would be great.

app/src/components/dashbaord/DashboardTask.tsx
erikguntner: @paulespinosa I'm not sure about the scope of the tasks for the MVP right now.

mira-kine (approved):
Feels clear to me in terms of data types and how it is organized. My only question is regarding rejection/withdraw events (per issue #391): would taskId just go back to host_matching again, and reset statuses? As in, do we need any way to keep track of the rejection/withdraw events?

agosmou (approved):
UI looks great! I'm working through the design of the endpoints, so I'll use some of this info to navigate the design.

Issue #580, opened 2023-08-27
API: Update connexion to 2.14.2, indirectly update Flask to 2.3.x series


paulespinosa:
### Overview
Update `connexion` to 2.14.2 and `Flask` to the 2.3.x series. Also, allow their transitive dependencies to be updated.

### Action Items

- [x] Review the change logs of [`connexion`](https://github.com/spec-first/connexion/releases) and [`Flask`](https://flask.palletsprojects.com/en/2.3.x/changes/)
- [x] Note the deprecations that happen when updating `Flask` from the **2.2.x** series to the **2.3.x** series
- [x] Update the API to address the deprecations and other necessary changes needed to be made to the API
- [x] Update connexion to 2.14.2. Doing this will indirectly update Flask to a version in the 2.3.x series. Allow for their transitive dependencies to be updated.
- [x] Test the API/Run the API

### Resources/Instructions
`connexion` release notes: https://github.com/spec-first/connexion/releases
`Flask` release notes: https://flask.palletsprojects.com/en/2.3.x/changes/

Joshua-Douglas:
Hey @johnwroge,

I think this would be a great first issue to work on! You'll need to build the backend and run the backend test cases to complete this issue. The [main README](https://github.com/hackforla/HomeUniteUs/blob/main/README.md) has docker build instructions, and the [api README](https://github.com/hackforla/HomeUniteUs/blob/main/api/README.md) has local build instructions. Building locally will be easier, but I think the README is less detailed. 

If you run into build error feel free to ping me or the dev slack channel. If you do find that the README is out of date, please update it as well!

johnwroge:
Hi Joshua! 

Thank you so much for the assignment and for the additional information! I looked over the ReadMe for the API directory and it's very well written. If I run into any issues with building and testing the api I will make a note and reach out. Paul sent over the .env file for to connect so I should be good to start this task today.

PR #601, opened 2023-09-19
Update connexion to 2.14.2, pytest to 7.4.2, and downgrade werkzeug to 2.2.3

johnwroge:
Closes https://github.com/hackforla/HomeUniteUs/issues/580

What changes did you make?
This PR includes the following changes:

· Removes the encoder.py file from the openapi_server directory and removes the encoder use in main, and init test files to eliminate test errors.
· Updates version of Connexion from 2.13.1 to 2.14.2
· Updates pytest from 7.3.1 to 7.4.2
· Downgrades Werkzeug from 2.3.7 to 2.2.3 to remove dependency conflicts

Rationale behind the changes?
The rationale behind these changes was to update Connexion and Flask to newer versions and to remove the deprecation warning the team was receiving when starting the virtual environment and running the test suite. Flask version 2.2.5 was unable to be updated due to conflicts with dependencies required for Connexion. Flask will have to wait to be updated until Connexion version 3.0 is released. The Werkzeug version had to be downgraded because Connexion version 2.14.2 requires version < 2.3 while the older version was compatible with 2.3.7.

 Pytest was updated because there were no breaking changes between versions according to the change logs, this version is still compatible with python 3.7+

https://docs.pytest.org/en/7.4.x/changelog.html

Testing done for these changes:
To confirm the test suite still passes, tests were run in the virtual environment in which 27 test cases passed.

What did you learn or can share that is new?(optional)
I learned through the process that when updating to later versions, it's important to check if there are any coexisting dependencies that require versions which are incompatible with each other.

Since this is my first PR please let me know if any additional information is needed for documentation.

### DIFF
--- a/api/openapi_server/__main__.py
+++ b/api/openapi_server/__main__.py
@@ -9,7 +9,6 @@ from dotenv import load_dotenv, find_dotenv, get_key
 import prance
 
 # Local
-from openapi_server import encoder
 from openapi_server.models.database import DataAccessLayer
 from openapi_server.exceptions import AuthError, handle_auth_error
 from configs.configs import compile_config
@@ -81,7 +80,6 @@ def create_app():
     # The underlying instance of Flask is stored in `connexion_app.app`.
     # This is an instance of `flask.Flask`.
     flask_app = connexion_app.app
-    flask_app.json_encoder = encoder.JSONEncoder
     flask_app.secret_key = SECRET_KEY
 
     # Below, the Flask configuration handler is loaded with the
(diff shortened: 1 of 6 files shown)


paulespinosa (approved):
@johnwroge great job! Thank you. :+1:

Joshua-Douglas (approved):
Hey @johnwroge, 

Thanks for this update! Approved. 

We had some discussions about warnings, so I took a moment to compare the warnings on this branch with `main`. This branch did not introduce any new warnings, and nearly all of the existing warnings appear to be originating from 3rd party packages (mostly `connexion`).

PR #622, opened 2023-10-09
Update host response

johnwroge:
Closes 583 - Bug[API] Controllers and OpenAPI response mismatch

What changes did you make?
I created a HostResponse returning an array of objects containing required key pairs of {name: string, id: integer}. Tests were written in Pytest to check that the controller was returning the correct information. The expected api response on the frontend was updated in Host.ts to reflect these changes.

Rationale behind the changes?
The host controller was previously set up to return a response of type api response containing code, type and message properties of types int32, string and string. These were different from what the host controller was actually responding so a new host response was created to match what should be returned.

Testing done for these changes
Additional unit/integration tests were written in the tests directory to check these changes. When running tox in a local environment, all the test cases passed. When running on the docker container, all the tests passed. There are a few warnings still being returned related to deprecations related to Connexion/Flask and how the current tests are written.

What did you learn or can share that is new?(optional)
I learned more about the open api spec and how the paths and responses are set up to return data in an organized format.


erikguntner (approved):
@johnwroge Thanks! This all looks good to me

paulespinosa (approved):

PR #628, opened 2023-10-25
Update build-deploy-ec2.yml

paulespinosa:
Automatically build and deploy code that reaches the main branch. Allow only one deployment job to run at a time because there are no other synchronization mechanisms in place.

  - Changed the build-deploy-ec2 workflow so that it runs on merge to main and PR to main as the target.
  
  - The api and app tests are now running on commits, merge to main, PR targeting the main branch. This gives some assurance that code can be deployed to the dev environment automatically without manual intervention. The caveat is when there are environment variable updates then the EC2 instance will need to be updated and the code manually deployed.
  
  - Can't test until it reaches the main branch.

  - GitHub workflows have concurrency controls to allow one workflow or job to run and one pending.


paulespinosa:
Incubator deployment is underway. This version will not be needed.

Issue #577, opened 2023-08-24
Introduce API Endpoint Authentication


Joshua-Douglas:
### Overview
Introduce secure API endpoint authentication, so that only registered users can access our API. The selected authentication approach should be compatible with our current use of AWS Cognito, and should be integrated with the frontend app.

### Action Items
- [ ] Research the available API authentication methods. Choose the approach that works best for our application and create a high-level for the required API changes
- [ ] Review design and chosen authentication method with team, to solicit feedback before beginning implementation
- [ ] Update our OpenAPI specification to reflect the required authentication method. Note that some endpoints, such as SignUp, will not be authenticated
- [ ] Use connexion's authentication system to enforce the authentication methods documented in the OpenAPI specification
- [ ] Design a system for managing developer and test credentials. We will likely need to either bypass authentication during development/testing, or provide a secure mechanism for easily requesting and storing developer credentials
- [ ] Update the backend and frontend test cases to utilize the test credential system
- [ ] Update the frontend app to integrate with the new authentication system. We will need to ensure that the client HTTP requests have the required authentication headers. This can be made easy using cookies or sessions.

### Resources/Instructions
* [AWS Cognito Documentation](https://docs.aws.amazon.com/cognitoidentity/latest/APIReference/Welcome.html)
* [Flask Security and Authentication](https://flask.palletsprojects.com/en/2.0.x/security/)
* [Connexion Authentication](https://connexion.readthedocs.io/en/latest/security.html)
* [Flask-login](https://flask-login.readthedocs.io/en/latest/)

Joshua-Douglas:
To ensure that the frontend app signin workflow continues to work I plan on introducing a suite of end-to-end integration tests that verify that our authentication works before and after these changes. 

I began writing the test cases, but those test cases discovered a bug in the `session` and `refresh` endpoints that needs to be addressed before the test cases can pass. 

This bug was documented in #633. My plan is to keep this issue in the ice box until #633 is completed, since:
1) this suite of test cases will be very valuable for this issue
2) I plan to make some useful refactors to the session and refresh endpoints in #633 that will make closing this issue easier

agosmou:
@Joshua-Douglas 
[Connexion 3.0 was released](https://connexion.readthedocs.io/en/3.0.2/security.html) - might be good to keep in mind if any of your implementation has the possibility to be depracated sooner with 2.0

Joshua-Douglas:
[...]

Issue #599, opened 2023-09-18
Consistent OpenAPI paths, parameters, and operationId naming convention


paulespinosa:
### Overview

The OpenAPI spec ([api/openapi_server/openapi/openapi.yaml](https://github.com/hackforla/HomeUniteUs/blob/main/api/openapi_server/openapi/openapi.yaml)) contains inconsistent naming conventions.

To promote consistent and predictable naming, align on a naming convention for OpenAPI parts (`paths`, `parameters`, and `operationId` and update the OpenAPI definitions, API code, and App code.

Across all naming conventions for parts, they should be:
* simple
* intuitive
* consistent

#### OpenAPI Paths
The OpenAPI spec uses at least two naming conventions for paths: mixedCase and snake_case.  The preference is to use all lowercase with hyphen separating words (e.g. /service-providers/).

#### OpenAPI and Python parameters

The OpenAPI parameters and Python function definition parameters should have consistent naming, preferably Python [PEP8](https://peps.python.org/pep-0008/#function-and-variable-names) snake_case.

- the parameters on `HomeUniteUs/api/openapi_server/openapi/parameters/_index.yaml`
- the parameters on paths `HomeUniteUs/api/openapi_server/openapi/openapi.yaml`
- the parameters on controller definitions

Documentation should be added about Python function definitions with parameters using Python built-in names should append an underscore to the paramater name. [Connexion parameter handling](https://connexion.readthedocs.io/en/latest/request.html#automatic-parameter-handling)

For example: `def example_func(filter_):` because `filter` is a Python built-in name for the [`filter()` function](https://docs.python.org/3.10/library/functions.html#filter).

#### OpenAPI operationId

The OpenAPI `operationId` should have consistent and predictable naming. The preference is to use Python [PEP8](https://peps.python.org/pep-0008/#function-and-variable-names) snake_case because operationId represents Python functions.

### Action Items

- [ ] Establish naming conventions (there will be different conventions used for paths, parameters, and operationIds)
- [ ] Update API: OpenAPI spec and Python code
- [ ] Update App: Calls to the API
- [ ] Tests
- [ ] Document naming conventions in README

### Resources/Instructions

Changes to paths in the API need corresponding changes to the app.

Initial conversation about naming conventions: https://github.com/hackforla/HomeUniteUs/pull/543#discussion_r1273005912
connexion documentation about parameter names: https://connexion.readthedocs.io/en/stable/routing.html#parameter-name-sanitation
[...]

PR #637, opened 2023-11-06
Fix configuration-specific CORS issues & cleanup session endpoints 

### LINKED ISSUE #633: Session and Refresh Endpoints Don't Work W/Default Local Configuration
### Overview

The backend API `auth/session` and `auth/refresh` endpoints always return a `401 UNAUTHORIZED` response when called from frontend app deployed locally using the default api and app configurations. Both of these endpoints rely on the session cookie to provide an updated JWT, but the session cookie is never attached to the backend API requests, so the authentication always fails. 

The major consequence of this bug is that the user is logged out anytime the web application is refreshed, or if the JWT expires and needs to be refreshed. 

Interestingly, this bug is not present on the dev.homeunite.us deployment. The missing session cookies are due to a local configuration issue.

### Root Cause

I've researched this issue and identified this as a [cross-origin resource sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) issue. By default our backend application is deployed to `127.0.0.1:8080` and our frontend application is deployed to `localhost:4040`. The frontend development server does forward requests made to `localhost:4040/api` to `localhost:8080`.

Our session cookies are not sent with this configuration because:
  1) `127.0.0.1` and `localhost` are considered as two different domains by browser [same-origin policies](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy) 
  2) same-origin policies prevent cookies from being shared between two different origins, unless the server enables CORS by setting the correct CORS headers in the cookie response. 

### Reproduction Steps

* Remove all unnecessary values from the `app/.env` and `api/.env` file (remote ROOT_URL and VITE_HUU_API_BASE_URL at least)
* Run the development API using default development configuration
* Run the frontend app using default development config (`npm run dev`)
* Sign-in
* Refresh page
* Notice that:
   1) you are logged out 
   2) the `refresh` & `session` requests failed with an auth error
   3) no session cookie was attached to the `refresh` and `session` endpoints
* Shut-down the frontend app
* Update `app/.env` to set `VITE_HUU_API_BASE_URL=http://127.0.0.1:4040/api/`
* Restart frontend app
* Sign-in
* Refresh page
* Notice that the refresh works now!

### Action Items
* Update the frontend/backend local build configuration to ensure that session cookies are attached to backend requests
* Update the `refresh` and `session` endpoints to return a custom status code if the session cookie is missing
* Write end-to-end integration tests to verify that refreshing the application and updating the JWT session works as expected
* (Optional) Review the session cookie settings to see if the proper security configurations are being used
* (Optional) Update the frontend vite development proxy to use a runtime value. The proxy is hard coded to the target `http://localhost:8080`, but the backend could be deployed on a different port
* (Optional) Explore ways to add an assertion to require that the frontend and backend application exist on the same domain. I may be able to achieve this by updating the session cookie configuration. If the two applications are served on different domains then we will need to update the backend to set the proper CORS headers. 
 
### Resources/Instructions
* [Flask session configuration options](https://flask.palletsprojects.com/en/2.3.x/config/#SESSION_COOKIE_NAME)
* [MDN HTTP cookie](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies)
* [MDN Cross Origin Resource Sharing](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS)

Joshua-Douglas:
Closes #633, and adds frontend/backend test suites that will be very useful for completing #577 (API Authentication).

This change ensures that user sessions work for all same-origin frontend and backend deployment configurations. 

Before this change the `session` and `refresh` endpoints would only work with our frontend app if the `VITE_HUU_API_BASE_URL` configuration values were set to exactly match the web application's base url.

|Web App Base Url| VITE_HUU_API_BASE_URL| Did Sessions Work?|
|---|---|---|
|127.0.0.1:4040|localhost:8080|No|
|localhost:4040|127.0.0.1:8080|No|
|127.0.0.1:4040|127.0.0.1:8080|Yes|
|localhost:4040|localhost:8080|Yes|

This misconfiguration was causing users to get logged out anytime the application reauthenticated the session or reloaded the page. The session endpoints failed to authenticate users because our session cookies were not being sent, due to browser cross-origin security policies. 

Despite the fact that `127.0.0.1` and `localhost` resolve to the same value in our case, browser [same-origin policies](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy) consider 127.0.0.1 and localhost as two different domains. Since our backend does not set CORS headers, these same-origin policies were preventing cookies from being shared with the backend. 

See #633 for a more in depth discussion of the root cause.

To fix this brittle configuration I updated the base query to use a relative API URL, which is always considered 'same-origin' by the browser. It is worth noting that we do use a proxy to forward the requests from the frontend to the backend, but same-origin policies do not apply to the proxy - they only apply within the browser. 

Here is the key fix:
```diff
// app/src/services/api.ts
const baseQuery = fetchBaseQuery({
-  baseUrl: import.meta.env.VITE_HUU_API_BASE_URL,
+ baseUrl: '/api',
```

### Other, Related Changes

While making these changes, I also refactored the frontend CORS related code and session endpoints. Most of these changes were driven by minor bugs discovered during testing.

- Correct the default API URL. Previously it was set to `http://localhost:4040/api/`, but the real location is `http://localhost:8080/api/`. 
- Update development server API proxy to work with any backend origin, by using the corrected API URL value to set the proxy target
- Remove unnecessary 'Access-Control-Allow-Origin' headers from the API requests. These headers have [no effect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) when added to requests. They are designed to be attached to responses.
- Correct session endpoint OpenAPI spec to apply JWT authentication. The JWT was being passed, but never used
- Correct the refresh & session endpoints OpenAPI specification to require a request session cookie, and to document the 401 error
- Correct signin endpoint OpenAPI spec to define the response json & header. Also, apply a format constraint on the email input and define the 401 error

#### Frontend

To test these changes I added a suite of `Cypress` e2e test cases that support a real and mocked backend. See the updated app `README` for full instructions on how to run the e2e test cases without mocking.

The e2e test cases verify that sessions work by ensuring that the user stays logged in during a page refresh, and by checking for the signin request cookies.

#### Backend

The backend test infrastructure does not support AWS Cognito mocking, so the suite of backend authentication tests are configured to only run in release mode. I focused on checking failure states since the frontend test suite covers the success cases. The plan is to eventually expand the backend suite to include successful logins, but this will require some additional modifications to our test infrastructure.


Joshua-Douglas:
Hey @erikguntner,

Can you try the staging configuration instead? AWSCognito features are temporarily disabled on the development configuration while we wait on mocking.

Bright side is that I have a branch with mocking enabled so we should be able to use the development configuration soon!

Let me know if you are still having issues while using staging.

erikguntner:
@Joshua-Douglas staging config worked! Thanks!

erikguntner (approved):
This looks great! I wonder if we can make some of the e2e test logic reusable for testing flows for authenticated users. Might be something we could create in a future PR.

app/.env.example
```diff
@@ -1,4 +1,4 @@
 SERVER_URL=http://localhost:6060
```
agosmou, line 1: What's `SERVER_URL` used for? Can this be deleted?

api/tests/test_authentication.py
```diff
@@ -0,0 +1,74 @@
+import string
+from tests import TestsWithMockingDisabled
+
+def strip_punctuation(text):
+    return text.translate(str.maketrans('', '', string.punctuation))
+
+class TestAuthenticationEndpoints(TestsWithMockingDisabled):
```
agosmou, line 7: Should we also test a working authentication that returns a 200 status?

If this was left off because a new db instance wont have seeded users, maybe we consider seeding a test user so we can run tests that return a '200' status. what do you think?

api/tests/__init__.py
```diff
@@ -53,3 +54,11 @@ def populate_test_database(self, num_entries) -> List[int]:
             ids.append(provider.id)
         return ids
 
+class TestsWithMockingDisabled(BaseTestCase):
```
agosmou, line 57: 👍🏽

api/openapi_server/controllers/auth_controller.py
```diff
@@ -13,6 +13,7 @@
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy import select
 
+from botocore.exceptions import ClientError
```
agosmou, line 16: using botocore errors for cleaner code?

api/openapi_server/controllers/auth_controller.py
```diff
@@ -167,13 +168,7 @@ def signUpCoordinator():  # noqa: E501
...
         raise AuthError({"message": msg}, 500)
 
-        
-
-def signin():
-    # Validate request data
-    if connexion.request.is_json:
-        body = connexion.request.get_json()
-
+def signin(body: dict):
```
agosmou, line 171: Should controllers be snake_case? This might be out of scope for this PR


[reference](https://github.com/hackforla/HomeUniteUs/issues/599)

agosmou:
Hi @Joshua-Douglas

I couldn't reproduce the error - Ill reach out async to discuss this!
___
Signed in as `testuser@email.com` and the opened a fresh (cleared) "Network" side-panel. I refreshed the page and a bunch of failed `refresh` and `session` popped up. How come this happens?
image

## Tests
### API
`api/tests/test_authentication.py` passed on my machine

### Cypress
`app/cypress/e2e/test-authentication.cy.ts` ran as `npm run cypress:open`
- chrome
    - image
- firefox
    - image

`app/cypress/e2e/test-authentication.cy.ts` ran as `npm run cypress:open:nomock`
- chrome
    - image

api/openapi_server/controllers/auth_controller.py
Joshua-Douglas: Good point. Since I touched it I'll fix the name. [6944201](https://github.com/hackforla/HomeUniteUs/pull/637/commits/69442013820dc721a022ab7c599456d9cddc8eb4)

api/openapi_server/controllers/auth_controller.py
Joshua-Douglas: Yes, because the old approach was error prone.

```python
    except Exception as e:
        code = e.response['Error']['Code']
```

We are catching all exceptions and then attempting to access a `response` property. `Exception` does not have a `response` property, so if an unrelated exception was thrown then `e.response` would cause another exception to occur.

api/tests/test_authentication.py
Joshua-Douglas: The tests in my next PR will include valid sign in and sign up attempts. I needed to do some more work to properly isolate test users from real users, to ensure that our test cases did not pollute our AWS data. 

See `https://github.com/hackforla/HomeUniteUs/blob/auth_mocking/api/tests/test_authentication.py` for example tests cases with `2xx` status. Note that the `TestsWithMockingDisabled` is no longer required and all API tests can be run with and without mocking in this upcoming PR.

app/.env.example
Joshua-Douglas: Good point. Deleted in [f232d57](https://github.com/hackforla/HomeUniteUs/pull/637/commits/f232d57183d9b59542a4beea66de09a44161fa05).

Joshua-Douglas:
Hey @agosmou, 

You outlined several items in your last post, so I'll address them separately. Thanks for the great questions! I had to do a bit of research to answer.

## test failures

I see the same test case keeps failing for you. I wasn't able to replicate on my machine, but it looks like a timing issue because the signin attempt succeeds but the url is incorrect. 

I updated the test case to include assertion retries (similar to the other test cases) in [99a35b1](https://github.com/hackforla/HomeUniteUs/pull/637/commits/99a35b141aa1ffdca4bacecba699ec75c7973ec0). Give it a try and let me know if it is still failing.

## Why the failed `session` and `refresh` calls during initial app load

This is by design. We use the API to check for a valid user session each time the app loads.

When the app loads it checks to see if the user is already logged in by querying the API `session` endpoint (see `main.tsx::HuuApp()`, it is called within the `useEffect`). The session query always fails because the API session cookies are currently designed to expire when the tab closes. If we decided to use persistent cookies in the future, then the session call here may succeed if the session cookie is stored in the browser.

`refresh` is called because of the logic we have in `api.ts::baseQueryWithReAuth`. This function defines the query logic that is used each time the API is queried. The logic here will attempt to refresh the user JWT anytime an authentication failure is encountered. 

I think the `session` failure is a good design and we should keep it. The logic within `baseQueryWithReAuth` is subject to debate, but outside the scope of this issue. For example we could either of these two alternative designs to reduce the number of failed queries:
1) Instead of refresh each time a `401` is encountered, we can update the API to return an 'expired token' error code and only refresh when that error code is encountered. 
2) We could design the API to automatically refresh expired JWTs if the request contains a valid refresh token. With this design the frontend wouldn't need to worry about token refreshes.

## Why the failed `user` & `session` calls during refresh?

This is also by design.

We store the user data and JWT in memory as a redux store. We store the refresh token as a session cookie. When the web app refreshes while the user is logged in a few relevant things occur:
1) The redux store is cleared (and the session cookie is not cleared)
2) The `HuuApp` `useEffect` will call the `session` endpoint
3) The `ProtectedRoute` will reauthenticate the user by calling the `user` endpoint

Since the JWT has been cleared the `user` endpoint query will fail until the JWT is refreshed using the session cookie. We could probably optimize this by checking for the JWT before attempting authenticated queries, but making this change would require significant changes since the authAPI doesn't have a way to distinguish between queries that require a JWT and those that don't. So I think this optimization is out of scope.

## Why is the `session` endpoint called twice each time?

The `session` is called within a React `useEffect`. We use `React.StrictMode`. Strict mode will automatically call `useEffect` functions twice for development builds as a tool to alert developers when the effect is not properly cleaning up after itself. 

So the duplicate session call will not occur in production builds. [Source](https://react.dev/reference/react/useEffect#my-effect-runs-twice-when-the-component-mounts)

Joshua-Douglas:
Hey @erikguntner, @agosmou, 

PR ready for re-review in response to your comments. I'll keep the e2e infrastructure in mind as I finish up my next two auth PRs! 

No rush on the review since December is our month off!

api/tests/test_authentication.py
agosmou: Those test are looking good! Thanks

agosmou:
> test failures

Having issues running cypress with my laptop (going to restart my laptop and try again). Code change looks promising! Used your changes as reference to familizarize myself with the differences based on cypress [docs](https://docs.cypress.io/api/commands/then#:~:text=then()%20allows%20you%20to,no%20assertions%20throw%20within%20it.)

> Why the failed session and refresh calls during initial app load?

Thanks for explaining the session and refresh logic. If this is by design, it's probably best to leave as-is for now, than to implement the alternative approaches. I can put some thought into this though - approach 1 being nice since we just edit the frontend, but approach 2 might be proper since it moves the load to the backend. hmm...

> Why the failed user & session calls during refresh?

Understood and agreed! Thanks

> Why is the session endpoint called twice each time?

Thank you for providing the docs for this. TIL!

agosmou (approved):

erikguntner:
@Joshua-Douglas  This all looks great to me! 

> I think the session failure is a good design and we should keep it. The logic within baseQueryWithReAuth is subject to debate, but outside the scope of this issue. For example we could either of these two alternative designs to reduce the number of failed queries:

> 1. Instead of refresh each time a 401 is encountered, we can update the API to return an 'expired token' error code and only refresh when that error code is encountered.

> 2. We could design the API to automatically refresh expired JWTs if the request contains a valid refresh token. With this design the frontend wouldn't need to worry about token refreshes.

I've been thinking about what you outlined in 1 as well. I think this would be a good update.

Merging!

PR #638, opened 2023-11-08
Upgrade node2023

### LINKED ISSUE #588: App: Node 16.x reaching end-of-life support
### Overview

Node 16.x is reaching End-Of-Life support on 2023-09-11. 

https://github.com/nodejs/release#release-schedule
https://nodejs.org/en/blog/announcements/nodejs16-eol 

### Action Items

- [ ] Update the app, packages, tools to the next version of Node (18.x for LTS or 20.x for current)
- [ ] Update documentation
- [ ] Update GitHub Actions

mira-kine:
Closes #588 

  - Followed the [general solutions plan](https://github.com/hackforla/HomeUniteUs/issues/588) on the issues page, except I upgraded to Node 20.x instead of Node 18.x. 
  - I created a pre-install check to enforce a minimum node version (20.x), set in the engine's section of package.json. When a new dev installs the node packages, a script will run to check the node version first per `check-node-version.mjs`. If the minimum node version is not met, the installation will fail and a console log error will show up requiring the current node version to be satisfied.
  - Added that minimum node version is enforced and how it works in README
  
  - The docker image currently runs on the latest version of node (as of October 18th, 2023, that is Node 20.x), so there was not as much to change there. The github work flow script is set to the lts of node instead of being specified as 20.x because we are always aiming to use the lts. 
  - For enforcing the minimum node version, I set it to >=20.0.0 because Node 20.x active support ends in 11 months and security support ends in 2 years [according to this schedule](https://endoflife.date/nodejs). Therefore I chose the lowest LTS version that will give us the longest time before security support ends. 
  - As for setting `esModuleInterop` in `tsconfig.json` as "true": EsNext module does not currently have a way to support package.json exports, which was necessary for my pre-install check file `check-node-version.mjs`. To make the least amount of changes, I chose to set esModuleInterop to true and change the file extension to mjs in order for the json import to work. I also used Node's `createRequire` to mimic CommonJS "require" statement, which allows package.json imports. 
 
  - I ran the build + ran the tests + manually tested the frontend before and after the upgrade
  - I changed the node version to a ridiculous number aka `"node": ">=50.0.0" ` and ran npm i to see if I would receive the expected error
  
  - I learned a lot about modules and Typescript compiler options regarding modules. I learned that TS needs to check that: 1. files are compiled into valid output module formats, 2. ensure that imports in the outputs (aka what is compiled out to runtime) will resolve successfully, and 3. what the type of the imports are. Other ways the package.json import could have worked was to change the `moduleResolution` to `nodenext` instead as a typescript compiler option, as NodeNext makes CJS and ESM coexistable, where as simple `Node` is meant only for cjs. Since we are using ESM however, the clash of interests made the package.json import harder than expected. I also could have used ts-node but I didn't want to keep adding dependencies for only one file.

<details>
<summary>Visuals after changes are applied, testing node minimum version</summary>
  
image

</details>


.github/workflows/build-deploy-ec2.yml
```diff
@@ -52,10 +52,10 @@ jobs:
         working-directory: ./app
     steps:
       - uses: actions/checkout@v3
-      - name: Use Node.js 16.x
+      - name: Use Node.js LTS
         uses: actions/setup-node@v3
         with:
-          node-version: "16.x"
+          node-version: 'lts/*'
```
agosmou, line 58: Is there any downside to writing `node-version: 'lts/*'` vs. specifying the LTS version e.g. `node-version: '20.x'`?

If the LTS version switches to 21, could our app break? Do we benefit from having more control (i.e. specificity)? You did a bunch of research so deferring to you!

app/package.json
```diff
@@ -31,10 +36,12 @@
     "@mui/material": "^5.8.6",
     "@reduxjs/toolkit": "^1.8.3",
     "formik": "^2.2.9",
+    "node": "^20.8.0",
```
agosmou, line 39: Is [this](https://github.com/hackforla/HomeUniteUs/pull/638/files#r1401513094) limited by line 39 here?

app/package.json
```diff
@@ -15,7 +18,9 @@
     "type-check": "npx tsc --noEmit",
     "prepare": "cd .. && husky install app/.husky",
     "pre-commit": "lint-staged",
-    "generate": "npx @rtk-query/codegen-openapi openapi-config.ts"
+    "generate": "npx @rtk-query/codegen-openapi openapi-config.ts",
+    "check-node-version": "node check-node-version.mjs",
+    "preinstall": "npm run check-node-version"
```
agosmou, line 23: What's the purpose of "preinstall"?

Also, should we be running the "check-node-version" in front of the build scripts to prevent the build if the developer's node is out of date?

app/check-node-version.mjs
agosmou, line 1: How come you use an mjs extension? Not familiar - curious!

Would it be beneficial to have a truthy conditional in an 'else' statement?
```javascript
if (!semver.satisfies(process.version, version)) {
  console.log(
    `Required node version ${version} not satisfied with current version ${process.version}.`,
  );
  process.exit(1);
} else {
  console.log(`Node version ${process.version} is ok.`); // added this line for truthy conditional
}
export { };
```

app/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "lib": ["es6", "DOM", "DOM.Iterable", "ESNext"],
     "allowJs": false,
     "skipLibCheck": true,
-    "esModuleInterop": false,
+    "esModuleInterop": true,
```
agosmou, line 8: guessing this has to do with using a .mjs file?

agosmou:
Hi @mira-kine ! This looks all good. I wrote some comments/questions - take a look!

Aside from that, just merge the main branch into your branch and resolve the merge conflicts, then see if it starts up ok in your environment.

app starts up just fine btw!

image

agosmou (changes requested):
left some comments for you!

app/package.json
erikguntner, line 39: Can we make node and semver dev dependencies since they aren't used in the app? Not sure on the size of the node package but might be best to keep it out of the production build.

Joshua-Douglas: @mira-kine, 

I believe this should be deleted. `node` is the runtime environment, not a packaged dependency. Specifying `node` here actually downloads [this package](https://www.npmjs.com/package/node) from the [node-bin-gen](https://github.com/aredridel/node-bin-gen) repository. It does not specify a `node.js` version. It looks like this repo is designed to install the LTS version of node and ignore the system node installs.

Joshua-Douglas: Agreed. We should delete `node` and move `semver` to the dev dependencies.

app/README.md
```diff
@@ -34,10 +36,10 @@ For production builds, vite preforms a string replacement on the `import.meta.en
...
+| ----------------------- | --------- | -------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `VITE_HUU_API_BASE_URL` | YES       | http://localhost:4040/api/ | The HUU API's base URL. In a development environment (mode is 'development' or 'test'): if this variable is not defined, then `http://localhost:4040/api/` will be used by default. In non-development environment: if this variable is not defined, then the build will throw an error. |
+|                         |           |                            |                                                                                                                                                                                                                                                                                          |
```
Joshua-Douglas, line 42: Optional, but can you revert this change? The extra spaces and `-` are ignored by the markdown renderer and there is an extra empty row now.

Joshua-Douglas (changes requested):
Hey @mira-kine, 

Great changes, and a creative approach for enforcing the node version. I didn't know about preinstall scripts. Very interesting! 

I'll approve once the `node` package dependency is removed. My other comments are optional. 

P.S. We are on break in December so feel free to wait until January to make changes!

app/README.md
mira-kine: Yes! I don't even know what happened here, good catch.

app/check-node-version.mjs
mira-kine: Hi @agosmou! I made this decision after looking at couple pieces of information: 1. I needed to create a check-node-version function in the same scope as package.json, because I would be needing to read it later on. 2. In order to create a function that enforces a minimum node version, I wanted to use SemVer, which is a semantic versioning specification. 3. In order to import SemVer, I wanted to adhere to the ES Module specification and use "import" because that is what the rest of our application is doing. However, the current version of Node defaults to CommonJS module if none is specified in package.json. This means that it expects a standardized module importing behavior using the keyword "require" instead of "import". Instead of adding `"type": "module"` to package.json which would be redundant and clash with our `tsconfig` file that specifies our app to use `ESNext` aka the most recent version of ESM, I just changed the extension to .mjs as an exception for that one file. 

It might look confusing because I also have to use "require" in the form of `createRequire` in this file because ESM currently cannot read json directly without it. I am basically explicitly defining a file as an ESModule standardized files while having to use CommonJS functionality within in order to work around what each can and cannot do. I hope that made sense and if not, please feel free to keep asking me questions/providing feedback for if you think this could have been done a cleaner way!

mira-kine: As for the truthy conditional, I like the suggestion so I will be adding it in my next commit. Thank you!

erikguntner:
@mira-kine It looks like eslint-plugin-prettier got installed at the root of the project, creating a package.json, package-lock.json, and node_modules folder. Maybe you accidentally npm installed at the root instead of the /app directory?

app/src/services/host.ts
```diff
@@ -4,6 +4,7 @@ const injectedRtkApi = api.injectEndpoints({
     getHosts: build.query<GetHostsApiResponse, void>({
       query: () => ({url: `/host`}),
       providesTags: (result = []) => [
+        // eslint-disable-next-line prettier/prettier
```
erikguntner, line 7: Can this be resolved?

mira-kine: yes!

PR #640, opened 2023-12-06
Guest application formik context #630

### LINKED ISSUE #630: Guest Application Stepper- Context Wrapper
### Overview
Create the Formik Context wrapper for the guest application form along with adding the application navbar button. This is apart of the broader "Stepper" feature of the guest application that tracks and guides the user through the application form. For this issue, buttons will only have navigation functionality and no submission functionality.

### Action Items
- [x] Add the application button to the Guest dashboard Navbar
- [x] Setup the route wrapper for the formik context container
- [x] Create the Context wrapper with the functionality tracking the data inputted in each form. 
- [x] Create the navigation buttons "Continue", "Back", etc. per Figma designs. Should only be for navigation purposes for now.
- [x] Test with dummy routes to make sure buttons work

### Resources/Instructions
[Broader scope for the application form](https://github.com/hackforla/HomeUniteUs/issues/610)
[Application Epic](https://github.com/hackforla/HomeUniteUs/issues/533)
[Figma Designs](https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/HUU-Everything-Figma-Nov-2022?type=design&node-id=16-2646&mode=design&t=1MJte1f7uXgDn5Ib-0)

JpadillaCoding:
Closes #630 

  - Made a context wrapper to track the data inputted on the form. 
  - Made the data Validation along with errors with Yup for each input. 
  - Added the context wrapper ready for use to add routes
  - Created Navigation buttons with the basic functionality of going to the next page, previous page, and exiting the application. 
  - Made a progress bar for showing the user where they are on the application.
  - Added the application to the sidebar nav for easy access in development. This will be changed later on. 
  - Context wrapper
    - Set up the initial values to empty since we are starting with a blank form. Currently, no data retrieval since we do not have a backend setup. 
    - The progress bar is currently hard coded to 10 pages per the Figma file. In the future, this should dynamically change when we know the designs are final. 
    - Current step is saved to local storage for the user to return to the current step 
    - Validation schema is dynamically rendered based on the state value of the step the user is on so each page is validated by itself
    - ``stepToRouteMapping`` is used to control which page is rendered. Each page that is added to the route should be added here for the stepper to know where to take the user. The name should match the path name in the ``main`` file.
    - Refer to ``PasswordResetContext`` in the ``main`` file as an example of how to add routes in the context wrapper.
 - Buttons
    - The current step is updated depending on the button. Since no backend logic is made yet the buttons only navigate but we intend to make these able to handle the data inputted.
- Validation
  - Most validators are pretty straight forward with what they require, which are strings. In this case, we just check that there is something inputted in the field along with an error message. 
  - Had to install a package for verifying dates due to complex edge cases along with formatting.
  - Used regex to accept valid US phone numbers. 
  - Made as an array to correspond with the pages being verified

  - Used basic boilerplate code to make 2 quick forms for testing the navigation, the context wrapper stepper, and also the dynamic stepper for validating one page at a time. 
 
<details>

N/A

</details>

<details>
  

https://github.com/hackforla/HomeUniteUs/assets/104275658/71676183-00bb-4c6f-bdb3-8c5e7efe0094

</details>


app/src/components/common/GuestApplicationButtons.tsx
```diff
@@ -0,0 +1,69 @@
+import {Button, Stack} from '@mui/material';
+import {useFormikContext} from 'formik';
+import {useNavigate} from 'react-router-dom';
+
+export const NavButtons = ({
```
erikguntner, line 5: Something I would consider doing is simplifying the props passed into this component. Instead of passing in `step, setStep, and stepToRouteMapping` we could move these handlers to the parent component and pass them in instead. That way we can pass in props like `next, prev, saveAndExit`. I think passing in these handlers instead of the pieces required for composing them is a little clearer to the developer and makes the components easier to test and refactor.

erikguntner (commented):
@JpadillaCoding this is looking great! I think this sets a solid foundation to build off of. I made some suggestions for the `GuestApplicationButtons` component so let me know what you think.

app/src/components/common/GuestApplicationButtons.tsx
JpadillaCoding: Made the changes Requested. Let me know what you think!

app/src/components/common/GuestApplicationContext.tsx
```diff
@@ -0,0 +1,134 @@
...
+  alcoholUse: '',
+  otherSubstanceUse: '',
+  mentalIllness: '',
+  programGoals: '',
+  hostRelationshipGoal: '',
+  potentialChallenges: '',
+};
+
+export const GuestApplicationContext = () => {
+  const TOTALPAGES = 10; /*Once all forms are made and added to "stepToRouteMapping", make this number change based on "stepToRouteMapping"*/
```
erikguntner, line 65: I'd make this TOTAL_PAGES since it's two words.

app/src/components/common/GuestApplicationContext.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const storedStep = localStorage.getItem('currentStep');
+  const initialStep = storedStep ? parseInt(storedStep, 10) : 1;
```
erikguntner, line 67: Is the 10 here in reference to the TOTAL_PAGES variable?

app/src/components/common/GuestApplicationContext.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const [step, setStep] = useState<number>(initialStep);
```
erikguntner, line 68: I think we can place lines 66 and 67 in the initializer function for the step state like this:
```
const [step, setStep] = useState<number>(() => {
   const storedStep = localStorage.getItem('currentStep');
   const initialStep = storedStep ? parseInt(storedStep, 10) : 1;
   return initialStep;
});
```
The way it's currently setup we're calculating the initial state and getting the step from local storage on every render. Here is some info from the [React docs](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state)

erikguntner (changes requested):
@JpadillaCoding This is looking good Jose. I have a couple suggestions regarding the setting of the initial step. This should be good to go after those changes are made

app/src/components/common/GuestApplicationButtons.tsx
erikguntner: Looks good! Thanks @JpadillaCoding

app/src/components/common/GuestApplicationContext.tsx
JpadillaCoding: No it is not, ``praseInt`` takes a secondary argument to know which type of base value  to convert the string to (binary, decimal, octo). Just happens that 10 is for decimal. 10 is the default value, but I explicitly added it since there can be edge cases that it defaults to one of the other options.

app/src/components/common/GuestApplicationContext.tsx
JpadillaCoding: Ahh, I see that I was initially accessing localStorage every render instead of only when it was needed which is the initial Render. Nice catch and I learned something new!

JpadillaCoding:
@erikguntner Requested changes have been made.

app/src/components/common/GuestApplicationContext.tsx
erikguntner: Ah thanks for clearing that up for me

erikguntner:
@JpadillaCoding Thanks for getting those changes in! Everything looks good to me.

PR #646, opened 2024-01-17
Fix user redirects and tests

### LINKED ISSUE #645: Update user redirects and tests
### Overview
Currently the user is redirected to the home page when signing out or trying to access a protected page and this needs to be updated to handle two cases:

1.  When a user signs out they are redirected to the home page
2. When a user tries to access a protected page they are redirected to the sign in page

Update the failing Cypress tests

erikguntner:
Closes #645 

  - Removed the resetting of credentials on logout
  - Updated the redirect when visiting protected outs from `/` to `/signin`
  - Removed the `/profile` path from the protected paths in the e2e tests since it no longer exists
  - Made `/signout` a protected route. I noticed it wasn't protected and therefore wasn't causing a token refresh if they were expired.

I made these changes to correctly redirect the user based on 2 cases:
1. When the user logs out they are redirected to `/`
2. When the user tries to access a protected page. For example, when they aren’t logged in or logged in as a different user role they are redirected to the `/signin` in page

One issue I ran into was the `ProtectedRoute` component was redirecting the user after logging out instead of allowing us to choose the path. This was being caused by this line of code in `app/src/components/common/AuthenticatedHeader.tsx`:

` appDispatch(setCredentials({user: null, token: null}))`

The `ProtectedRoute` component relies on the user object to determine whether to navigate the user to `/signin` or render the necessary components. A change to the user object was causing the component to render`<Navigate to="/signin" state={{from: location}} replace />;` and override our call to `navigate('/')`. By removing the state update we can now redirect the user to the home page. However, the user object is still stored in state, but since the credentials are no longer valid the user can't log in or access any protected routes. The user object will also be reset anytime the user revisits the application or a protected endpoint.  

  - Updated Cypress tests to reflect the proper route redirects. There were also a lot of linter updates that were updated as well.

<details>

image

</details>

<details>
  
https://github.com/hackforla/HomeUniteUs/assets/27253583/68faddde-ba1c-4c92-97cf-924527676abc

</details>


JpadillaCoding (approved):
The intended changes of redirecting the user on logout to ``/`` works and the user is also redirected to ``/signin`` when they are not logged in and try to access a protected route. 

Thanks for the fix Erik!

PR #665, opened 2024-03-13
Fix mocking errors between Node v16 and Node 18+, clean up repository

erikguntner:
  - Upgrade MSW to version 2 and refactor mock handlers
  - Refactor tests that use mock handlers
  - Update the base URL for API calls to use the base URL from the environment
  - upgrade actions and Node versions in CI since MSW v2 requires Node 18+ 
  - The integration and unit tests using MSW were failing on local machines since the base URL wasn't recognizable in test environments.


JpadillaCoding:
Hey Erik I was reviewing the changes but am having a hard time understanding what the issue was since I never experienced it myself. Could you help me by telling me how I can mimic the issue? 
I also noticed that up until this point I never had a .env file for the front-end. Will this be necessary for this issue integration? I see that ``VITE_HUU_API_BASE_URL`` is used here to address the issue. If so let me know what I would need to setup on that file if anything differs from the .env example.
Thanks!

Joshua-Douglas:
Hey @erikguntner, I'm running into the same issue as @JpadillaCoding. Can you provide some reproduction steps using the `main` branch?

I just performed the following steps locally on `main` using node v16 on windows and the tests passed in each case.
1. `npm ci && npm run test`
2.  `npm run dev` (one terminal session), `cypress run` (another terminal session)
3.  `python -m openapi_server` (one terminal session), `npm run dev` (another terminal session), `$env:CYPRESS_REAL_EMAIL="test@test.com" && $env:CYPRESS_REAL_PASSWORD="Test!123" && cypress run --env USE_MOCK=false` (a third terminal session)

Also the 1) and 2) examples are running on our GitHub Action runner in linux without issue.

It is probably worth noting that I ran each test case without an `.env` file present. It is possible that the `.env` file may interfere with our tests.

Joshua-Douglas (changes requested):
Hey @erikguntner, I'm running into the same issue as @JpadillaCoding. Can you provide some reproduction steps using the `main` branch?

I just performed the following steps locally on `main` using node v16 on windows and the tests passed in each case.
1. `npm ci && npm run test`
2.  `npm run dev` (one terminal session), `cypress run` (another terminal session)
3.  `python -m openapi_server` (one terminal session), `npm run dev` (another terminal session), `$env:CYPRESS_REAL_EMAIL="test@test.com" && $env:CYPRESS_REAL_PASSWORD="Test!123" && cypress run --env USE_MOCK=false` (a third terminal session)

Also the 1) and 2) examples are running on our GitHub Action runner in linux without issue.

It is probably worth noting that I ran each test case without an `.env` file present. It is possible that the `.env` file may interfere with our tests.

erikguntner:
@Joshua-Douglas @JpadillaCoding I made some new updates to this that I think cleaned up some of the issues with the original changes.

- I removed any usage of `import.meta.env.VITE_HUU_API_BASE_URL` that I had previously implemented and changed the API baseUrl to be `new URL('/api', location.origin).href`. You can see the changes in commit [df48a93](https://github.com/hackforla/HomeUniteUs/pull/665/commits/df48a938bfd11fcdeb9b80327e32ccdf5b305145). These changes were made as suggested in this [issue](https://github.com/reduxjs/redux-toolkit/issues/3284) and the [MSW docs](https://mswjs.io/docs/runbook#rtk-query-requests-are-not-intercepted). They go into more detail about the issue with relative paths and base URLs in testing environments.
- Upgraded Vite, Vitest, @vitejs/plugin-react to the latest versions.
- Removed Storybook. I ran into dependency conflicts with Storybook packages while upgrading Vite, and since Storybook isn't being used, I decided to remove it from the project instead of upgrading the related packages.
- Removed the `/data` and `/model` as they were no longer being used either. 

Please let me know if you have any further questions about the updates. Thanks!

JpadillaCoding:
Hey @erikguntner  sorry for the late response. 
I've pulled your changes and performed npm install.
I'm using node version 21.7.1 and every test is passing except for ``ForgotPassCode.test.tsx``. 
Oddly enough when I use v16.0.0 I get 9 file failures when Joshua is having them all pass. 
When using v20 I get them all to pass no problem.

erikguntner:
>I'm using node version 21.7.1 and every test is passing except for ForgotPassCode.test.tsx.
Hmm could you provide more info for this? Might be a Node v21 issue?

> Oddly enough when I use v16.0.0 I get 9 file failures when Joshua is having them all pass.
This should be the case since MSW v2 doesn't support  v16. I think they were passing for Josh since he was still on `main` and hadn't pulled in the changes yet. @Joshua-Douglas is that right?

> When using v20 or v18 I get them all to pass no problem, so the node migration should be helpful for this issue.
That's the ultimate goal. Since v16 is coming to its end of life, these package upgrades will get us off of v16 for the most part and make us compatible with node v18+.

JpadillaCoding (approved):

PR #667, opened 2024-03-13
removed push: from github actions to only run on pull requests instead

### LINKED ISSUE #666: Remove push from current GitHub actions
### Overview
Currently, GitHub actions runs api and app tests on push and pull. We only want it to run during pull because when they run simultaneously, they end up running both twice. Fixing this issue should also fix the currently tests that were failing on push.

### Action Items
- [ ] remove push action from current github actions yml file

mira-kine:
Closes #666 

  - Removed running tests on push in gitHub actions 

  - GitHub did not run any tests on push when I pushed my commit

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,6 +1,5 @@
 name: Run API and App Tests
 on:
-  push:
   pull_request:
     branches: [main]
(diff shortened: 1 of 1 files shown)


erikguntner (approved):
Looks good to me!

PR #661, opened 2024-02-21
Collect first and last names

erikguntner:
Closes #660 

### What changes did you make?
  - Added fields for first and last name to sign up form
  - Update validation schema for first and last name
  - Updated requests to `signup/host` and `signup/coordinator` to include firstName and lastName properties
  - Updated request types for services
  - Wrote unit tests for sign up form and e2e tests testing happy path of signing up
  - The only changes made to `app/cypress/e2e/test-authentication.cy.ts` were solely prettier formatting
### Rationale behind the changes?
  - We need to collect first and last names when a user is created in order to properly display their names
### Testing done for these changes
  - Wrote unit and e2e tests

### Screenshots of Proposed Changes Of The Website  (if any, please do not screen shot code changes)

<details>
<summary>Visuals before changes are applied</summary>

</details>

<details>
<summary>Visuals after changes are applied</summary>
  
Screenshot 2024-02-20 at 6 28 24 PM

</details>

PR #602, opened 2023-09-19
NavMobile landing page side bar links

mira-kine:
Closes #579 

### What changes did you make?
  - Added href to list item button in the drawer as well

### Rationale behind the changes?
  - current link was only used in the HUU Logo button, not the hamburger menu drawer
  
### What did you learn or can share that is new?(optional)
  - coordinator/host icon responsiveness will be changed in another issue
  - nav mobile has both hamburger and HUU icon working as nav (maybe a merge bug?). Since issue #552 was about replacing hamburger menu with HUU icon, I think this is another bug but wanted to make sure since I see a [double menu proposal on Figma](https://www.figma.com/file/BNWqZk8SHKbtN1nw8BB7VM/HUU-Everything-Figma-Nov-2022?node-id=370%3A0&mode=dev) as well.

<details>
<summary>Visuals before changes are applied</summary>

[Original Bug](https://hackforla.slack.com/files/UJK101BKL/F05PV6ELW3B/huu-mobilenavbarbug.mp4)

</details>

<details>
<summary>Visuals after changes are applied</summary>
 
https://github.com/hackforla/HomeUniteUs/assets/90009901/fc1cebf8-7e8b-4d56-8cdf-133422bf251c

</details>

PR #668, opened 2024-03-24
655 Add User Roles and Basic User Data to Db

### LINKED ISSUE #655: Update User Database Model to Include User Types
### Overview

The current user model does not differentiate between user types, and only includes the username and email. To implement the coordinator dashboard requirements, outlined in #501 and #651, we need to make the following changes to our database:

* Update the `User` model to include first and last name
* Update the `User` model to include account type (`Guest`, `Host`, `Coordinator`, `Admin`)
* Update `Guest`, `Host` and `Coordinator` models to allow for the following relationships:
    * Coordinators can have 0 or more Guests
    * Coordinators can have 0 or more Hosts
    * Guests can have 0 or 1 Coordinators
    * Guests can have 0 or 1 Hosts
    * Hosts can have 0 or more Guests
    * Hosts can have 0 or 1 Coordinators

To facilitate these changes we will need to update the `SignUp` endpoints to specify the role type, first name, and last name. None of our endpoints support creating an association between users, so we will need to add this capability in a future issue.

### Action Items
* (Required) Update User models to include user types, and to define the user relationships
* (Required) Write test cases to verify that we can store and retrieve users from the database
* (Required) Update the SignUp endpoints to include basic user info, and the user type
* (Optional) Write a script that uses faker to populate a test database with these relationships in place. We already have a collection of test users. It would be useful to load useful data for these users.
* (Optional) Add middleware to validate role-based access. In general guests and hosts should only be able to access their own info. Coordinators have elevated access.
* (Optional) Add endpoints to query the user assignments. We could do these:

#### Optional Assignment Endpoints

1. Get a Coordinator's Assigned Guests and Hosts
* Endpoint: /coordinators/{coordinator_id}/assignments
* Method: GET
* Description: Retrieves both the guests and hosts assigned to a specific coordinator. The response could be structured to include separate lists for guests and hosts.

```
{
  "coordinator_id": 1,
  "guests": [
    {
      "user_id": 2,
      "first_name": "John",
      "last_name": "Doe",
      "host_id": 4
    },
    {
      "user_id": 3,
      "first_name": "Jane",
      "last_name": "Doe",
      "host_id": 5
    }
  ],
  "hosts": [
    {
      "user_id": 4,
      "first_name": "Host",
      "last_name": "One"
    },
    {
      "user_id": 5,
      "first_name": "Host",
      "last_name": "Two"
    }
  ]
}
```

2. Get a Guest's Coordinator and Host
* Endpoint: /guests/{guest_id}
* Method: GET
* Description: Retrieves details of a specific guest, including their assigned coordinator and host.
```
{
  "user_id": 2,
  "first_name": "John",
  "last_name": "Doe",
  "coordinator": {
    "user_id": 1,
    "first_name": "Coordinator",
    "last_name": "One"
  },
  "host": {
    "user_id": 4,
    "first_name": "Host",
    "last_name": "One"
  }
}
```
3. Get a Host's Coordinator and List of Guests
* Endpoint: /hosts/{host_id}
* Method: GET
* Description: Retrieves details of a specific host, including their coordinator and a list of assigned guests.

```
{
  "user_id": 4,
  "first_name": "Host",
  "last_name": "One",
  "coordinator": {
    "user_id": 1,
    "first_name": "Coordinator",
    "last_name": "One"
  },
  "guests": [
    {
      "user_id": 2,
      "first_name": "John",
      "last_name": "Doe"
    },
    {
      "user_id": 3,
      "first_name": "Jane",
      "last_name": "Doe"
    }
  ]
}
```

Joshua-Douglas:
Closes #655.

Added user roles (`Guest`, `Host`, `Coordinator`, `Admin`) and updated the user model to include first, middle, and last name. PR #661 made an update that passes the first and last name to the backend, so I was also able to update the frontend app to begin receiving and storing the user names as well (video below).

The user name initials are now displayed on the user avatar menu after logging in, and the user's name is accessible from all authenticated components.

Making this change required refactoring our alembic migration scripts, the API test project, and the API itself. The following detailed changes were required:
- Add a database version check during API startup. If the database is not at the head revision then an error will be thrown to notify the developer that an alembic migration is required. The logic relies on a static version number that should be incremented.
- Integrate the alembic database migration process into our test client setup fixture. Our backend tests will run against a properly versioned database, to ensure the test database has all pre-populated values (such as the `roles` table). 
- Upgrade `marshmallow-sqlalchemy` to the latest release. I needed a bug fix from the latest release.
- Update `signUpHost` and `signUpCoordinator` endpoints to 1) assign a role to their created users and 2) require the user first and last name.
- Update the `user` and `signIn` endpoints to return the user first and last name as part of the `user` data.
- Update data model diagram within `api/openapi_server/models/README.md`, and include instructions on how to update the database model. Turns out there are a lot of steps!
- Add `User` model (de)serialization json schema, complete with a test suite. We could add some name validation here (e.g. don't allow spaces in names) and filter the json requests through the schema, but it didn't seem worth while.
- Remove the `Host` model object. This is now replaced by the `User` model with a `Host` role.
- Remove the `create_host` endpoint. This is now completely replaced by `signUpHost`
- Modify the base alembic migration script, `3ceec084158f_.py`, to work with unversioned local databases. This will allow other devs to run `alembic upgrade head` even if they've been working off of an API-generated database.
- Add migration script to introduce user roles and user names. All existing users will be assigned the `guest` role and will be given the names "unknown".
- Update `run-tests.yml` to include a `workflow_dispatch` trigger to allow manual triggering.
- Add `cypress:run:mock` and `cypress:run:nomock` configurations to allow more easily running the full e2e test suites.

Before this PR there was no mechanism for determining user type and there was no mechanism for storing basic user information. We can now differentiate between guests, hosts, and coordinators. This will allow us to begin creating different views for each user type, and to begin restricting portions of the API that require elevated access.

This PR does not implement any data model relationships between user types, and it does not implement role-specific user data.

  - Added a suite of backend alembic migration tests that will test all current and future migration scripts. These tests cycle through the upgrades and downgrade scripts and check for common database migration errors. These tests are assisted by the new dev dependency, `pytest-alembic`.
  - Modified the e2e sign up tests to include a 'no mocking' mode. I ran the signup tests against the real API to confirm our signup still works.
  - Tested each of the modified endpoints
  - Tested the new schema objects, and their validation

I learned how to write and test `alembic` database migrations. The steps are documented in `api/openapi_server/models/README.md`. 

### New Data Model
```mermaid
classDiagram
class alembic_version{
   *VARCHAR<32> version_num NOT NULL
}
class housing_program{
   *INTEGER id NOT NULL
   VARCHAR program_name NOT NULL
   INTEGER service_provider NOT NULL
}
class housing_program_service_provider{
   *INTEGER id NOT NULL
   VARCHAR provider_name NOT NULL
}
class role{
   *INTEGER id NOT NULL
   VARCHAR name NOT NULL
}
class user{
   *INTEGER id NOT NULL
   VARCHAR email NOT NULL
   VARCHAR<255> firstName NOT NULL
   VARCHAR<255> lastName NOT NULL
   VARCHAR<255> middleName
   INTEGER role_id NOT NULL
}
housing_program_service_provider "1" -- "0..n" housing_program
role "1" -- "0..n" user
```

### Video of new Sign In User Initials
<details>
 
https://github.com/hackforla/HomeUniteUs/assets/22138019/9bdf0326-269e-432b-be1f-335f64605367

image

</details>


.github/workflows/run-tests.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches: [main]
   # This is also a reusable workflow that can be called from other workflows
   workflow_call:
+  workflow_dispatch:
```
erikguntner, line 7: What is this doing?

erikguntner (approved):
Everything looks great! Thanks for the detailed steps for updating the data model. Approving. Only question is regarding the `workflow_dispatch` property in the test workflow.

api/openapi_server/models/database.py
```diff
@@ -1,14 +1,26 @@
...
 
 class User(Base):
     __tablename__ = "user"
     id = Column(Integer, primary_key=True, index=True)
     email = Column(String, nullable=False, unique=True)
+    firstName = Column(String(255), nullable=False)
+    middleName = Column(String(255), nullable=True)
+    lastName = Column(String(255), nullable=False)
+    role_id = Column(Integer, ForeignKey('role.id'), nullable=False)
+    role = relationship("Role", back_populates="users")
```
erikguntner, line 17: Does this return the role as a string on the user object?

api/alembic/versions/e4c8bb426528_add_user_types.py
```diff
@@ -0,0 +1,70 @@
...
+        1. role - Store available application user roles
+    2. Prepopulate the role table with four role types: Admin, Host, Guest, Coordinator
+    3. Update the user table to add the first, middle, last name, and role_id columns.
+      * All existing users will be given the first, last name "UNKNOWN"
+      * Assign all existing users to the Guest role.
+    4. Drop the host table. 
+      * There is no way to map host users back to the user table. We would need a user id foreign 
+        key, or at least an email address.
+    '''
+    role_table = op.create_table('role',
```
paulespinosa, line 30: This means there can be duplicate role names. The `id` column isn't strictly necessary. The `name` column can be the primary key.

paulespinosa (changes requested):

paulespinosa:
> * (e.g. don't allow spaces in names)

Some people have spaces in their names. Here's some resources to understand more about names:

* https://designsystem.digital.gov/patterns/create-a-user-profile/name/
* https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/

https://github.com/kdeldycke/awesome-falsehood

paulespinosa:
Looking good. Is there a PM created issue this can relate back to?

.github/workflows/run-tests.yml
Joshua-Douglas: The `workflow_dispatch` event lets you manually trigger the workflow. Sometimes it is useful to run the remote test cases before submitting a PR, so I usually like to include this.

api/openapi_server/models/database.py
Joshua-Douglas: It returns a `models.database.Role` object. The `relationship` function uses the `role_id` to perform an automatic lookup of the `Role` object whenever the `user` object is accessed. It'll also perform some validation when updating the `role_id` to make sure it is a valid `Role`. 

The `User` table doesn't actually store a `role` - just the `role_id`: 
image

But if you dump the `user` model to json you see the serialized `role` object:
`{'role': {'name': 'Admin'}, 'email': 'testcoordinator@test.com', 'firstName': 'testuser_firstname', 'middleName': '', 'lastName': 'testuser_lastname'}`

erikguntner: Okay cool. If that's the object being returned when the user logs in then that's perfect. Just wanna make sure we have the role on the front end so we can do things like render routes based on role, show different components, etc.

Joshua-Douglas:
> > * (e.g. don't allow spaces in names)
> 
> Some people have spaces in their names. Here's some resources to understand more about names:
> 
> * https://designsystem.digital.gov/patterns/create-a-user-profile/name/
> * https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/
> 
> https://github.com/kdeldycke/awesome-falsehood

That's a great illustration of why I did not validate the names. Great sources, thanks!

After reading through them I realized two things and corrected both in [6ad02dd](https://github.com/hackforla/HomeUniteUs/pull/668/commits/6ad02dd8234cceeb29d77c55fff3045dfaf495c8):
1. Last names should not be required. I updated the database model to only require first names.
2. First names should be at least one character in length. The data model now validates that the name is at least one non-space character.

api/alembic/versions/e4c8bb426528_add_user_types.py
Joshua-Douglas: Correct, the `id` isn't strictly necessary, but it does provide us the flexibility to rename the roles at any time. I think being able to easily rename our roles is worth storing 4-10 extra integers in the db.

To your point about duplicate names, I added a safeguard against this in [6ad02dd](https://github.com/hackforla/HomeUniteUs/pull/668/commits/6ad02dd8234cceeb29d77c55fff3045dfaf495c8) by adding the `unique=True` constraint to the `role.name` field.

Joshua-Douglas:
Hey @erikguntner and @paulespinosa, 

Thank you so much for taking the time to do a detailed review. I've updated the PR in response to your comments. 

Since I had to modify the migration script, it is probably best for you to downgrade and re-upgrade if your local db is on the current head. 

Here are all the changes from the review:
- Make the user last name optional, since some cultures only use one name. I modified both the frontend and backend to accommodate this change.
- Add validation to the user first name, to require at least 1 non-space character in length.
- Auto-increment the database head version by reading from the alembic configuration file. Before this change the version had to be manually incremented, but this change removes a manual step from the db upgrade process.

I also noticed that #602 broke our e2e tests, so I submitted a fix for this in [74f861e](https://github.com/hackforla/HomeUniteUs/pull/668/commits/74f861e4e8bf16b52a9cc5cffa2e0607e23cba9c).

api/alembic/versions/e4c8bb426528_add_user_types.py
paulespinosa: Using `on update cascade` can also be considered easy when defining a table. What other reasons are there to use a surrogate key?

Joshua-Douglas: Using surrogate keys like an id field offers several advantages beyond the ease of updates provided by `ON UPDATE CASCADE`. Surrogate keys are unique and stable, providing an abstraction layer that decouples database structure from business logic, thus reducing the need for changes due to business modifications. They enhance performance with quicker indexing and simplify relational management by avoiding the complexities and potential performance issues of cascading updates. Additionally, surrogate keys maintain data integrity by minimizing risks associated with data duplication and erroneous updates, making them ideal for robust, scalable database architectures.

Joshua-Douglas:
@paulespinosa, I saw your question but I don't see any requested changes. Does this PR require more work?

api/alembic/versions/e4c8bb426528_add_user_types.py
paulespinosa: Is that a chatgpt response? :smile: Do these reasons apply to this data model?

paulespinosa:
> @paulespinosa, I saw your question but I don't see any requested changes. Does this PR require more work?

Code seems fine. @sanya301 this feature adds fields to the sign up page as shown in the video above. Look good to you?

paulespinosa (approved):

PR #671, opened 2024-03-30
565 signup failure rollback

### LINKED ISSUE #565: Backend Sign Up Methods do not rollback changes during authentication failure
### Overview

We currently authenticate users using AWS Cognito, within the `signUpHost` and `signUpCoordinator` methods contained within `/api/openapi_server/controllers/auth_controller.py`. 

https://github.com/hackforla/HomeUniteUs/blob/47d7dc6e968b7881d6953c723991e8f12f1ad002/api/openapi_server/controllers/auth_controller.py#L130-L170

The signup method has two basic steps, that both have the possibility of failing:
1) Update the database with the new user email
2) Make an API request to AWS Congito to sign up the user

The problem is that if step 1 succeeds but step 2 fails, then the user database will contain a user email that was not added to the AWS Cognito database. If the signup fails then we should rollback change to our user database, in order to keep the AWS Cognito user entries and postgres database user entries synchronized.

Besides introducing a design flaw, this can also prevent users from signing up. If user re-tries the signup endpoint then the attempt will always fail because the user email is already present in the postgres user database.

### Action Items
Update the `auth_controller` signup methods to provide atomic updates to the AWS cognito and local database. If either update fails, then both updates should be rolled back.

JpadillaCoding:
Closes #565 

  - Added a rollback feature of removing the user from the database in case Cognito experienced a failure 
  - The database and Cognito would be out of sync in the case of Cognito experiencing a failure. 
  - Three approaches were possible:
    - Adding a deletion of the email if Cognito failed. 
    - Changing the sequence to Cognito being completed first and updating the databases second, but this option still had the issue of having to rollback Cognito if the database request failed. I found it easier to reverse a database entry than reverse Cognito.
    - Embedding the Cognito request within the try and except blocks of the database entry so the built in ``session.rollback`` method could be used. I found that this made the code harder to read since it added more nesting of try and except blocks.
  - I went with the first approach since it was the simplest and easiest to read. I made the deletion logic its own function for reusability. 
  - The cases for the Cognito Errors to rollback the database entry were kept for all cases except if the username already exists within Cognito. I checked each possible error and found this case to be the only one where an email should not get deleted since it signifies the user exists. Each case can be found [here](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_SignUp.html) in the ``Errors`` section
  - Manually tested Cognito failure by hard coding the try blocks password payload to an invalid password
`Password=body['password']` ---> `Password='lol'` 
This invoked the ``'InvalidPasswordException'`` case and verified the email deletion worked. 

### What did you learn or can share that is new?
  - Amazon documentation is vast and multiple links can have relevant information for your use case. It's good to ask around to make sure you did not miss anything 

<details>

image

</details>

<details>
  
image

</details>


api/openapi_server/controllers/auth_controller.py
```diff
@@ -58,6 +58,22 @@ def get_token_auth_header():
...
+            }, 404)
+
+        try:
+            session.delete(user)
+            session.commit()
+        except IntegrityError:
+            session.rollback()
+            raise AuthError({
+                "message": "Failed to delete user"
+            }, 500)
```
Joshua-Douglas, line 76: I added a very similar method that removes users from the db and AWS Cognito in #641, `admin_controller.removeUser`. You could almost use it here, but you'd have to handle the `AuthError` that is thrown when the user is not found in the AWS Cognito pool. 

This is optional, but it would reduce duplicate logic if you modify `removeUser` to avoid re-raising the `UserNotFoundException` `AuthError` and reusing the logic here. (also I should have named the function `remove_user` 😅).

Joshua-Douglas (changes requested):
Hey @JpadillaCoding, 

Great change! I just have an optional suggestion, and I'd like to request a regression test to ensure we don't run into this issue again. 

You asked over slack how to do this so I thought I'd write up an example test you could use.

This test should fail before your fix and pass after your fix. If you can think of other signup inputs that would trigger the bug, then feel free to add them.

```py
def test_user_signup_rollback(app):
    '''
    Verify that a failed signup attempt does not update the 
    user database or the AWS Cognito userpool. 

    This is a regression test. Previous implementations of 
    signup failed to rollback local db changes when the 
    AWS Cognito signup failed.
    '''
    rollback_email = 'test_user_signup_rollback@fake.com'
    signup_response = app.test_client().post(
        '/api/auth/signup/host',
        json = {
            'email': rollback_email,
            'password': 'weak'
        }
    )
    assert signup_response.status_code == 400
    with pytest.raises(app.boto_client.exceptions.UserNotFoundException):
        app.boto_client.admin_delete_user(
            UserPoolId=app.config['COGNITO_USER_POOL_ID'],
            Username=rollback_email
        )
    with DataAccessLayer.session() as sess:
        rolledback_user = sess.query(User).filter_by(email=rollback_email).first()\
        # This assertion will fail on `main` because no rollback is happening
        assert rolledback_user is None
```

api/openapi_server/controllers/auth_controller.py
JpadillaCoding: Great thanks for providing some insights on how to accomplish this 😄

JpadillaCoding:
@Joshua-Douglas I took your advice and used remove_user to avoid duplicate logic. To avoid the errors you mentioned and to also make the function more flexible for future uses, I made the deletion from cognito and the DB separate by specifying which to do on the parameters. I also set deleting from both as default as most use cases would call for the deletion of both. Let me know what you think of this approach. 

As far as the test you provided, it worked fine as is and didn't need any modifications. The test failed without remove_user implemented on the case errors and passed after it was implemented. 

Thanks!

api/tests/test_authentication.py
```diff
@@ -252,3 +253,27 @@ def test_session_endpoint(client):
...
+        }
+    )
+    assert signup_response.status_code == 400
+    with pytest.raises(app.boto_client.exceptions.UserNotFoundException):
+        app.boto_client.admin_delete_user(
+            UserPoolId=app.config['COGNITO_USER_POOL_ID'],
+            Username=rollback_email
+        )
+    with DataAccessLayer.session() as sess:
+        rolledback_user = sess.query(User).filter_by(email=rollback_email).first()\
```
tylerthome, line 277: possible typo on the trailing `\` here

api/openapi_server/controllers/admin_controller.py
```diff
@@ -55,7 +55,7 @@ def initial_sign_in_reset_password():
         'user': user
     }
 
-def removeUser(body: dict):
+def remove_user(body: dict, removeDB: bool = True, removeCognito: bool = True):
```
tylerthome, line 58: Since all of this method's logic is nested under `if` statements driven by the caller's params, can we log a warning and/or error out if both are passed as false?

Generally speaking, I think it would be nice for some helper methods here for clarity, e.g. `remove_user_from_db(body:dict[str, str])` and `remove_user_from_cognito(body:dict[str, str])` to be called from this method's internals, although this works fine and is still very readable as is currently.

api/openapi_server/controllers/auth_controller.py
```diff
@@ -96,15 +98,19 @@ def sign_up(body: dict):
                 raise AuthError({  "message": msg }, 400)
             case 'InvalidPasswordException':
                 msg = "Password did not conform with policy"
+                remove_user(body, True, False)
```
tylerthome, line 101: since we are using bool args for this method's params, these calls would be more readable as e.g.
```python
remove_user(body, removeDB=True, removeCognito=False)
```

An alternative design could be a integer enum flag, so we could pass one, neither or a binary `OR` as below:
```python
remove_user(body, RemoveUserOptions.RemoveFromDb | RemoveUserOptions.RemoveFromCognito)
```

tylerthome (changes requested):
This is looking good to me so far @JpadillaCoding , I added a couple of notes here. The only hold-up for me is to confirm whether trailing backslash is intended as it will escape the newline character and yield unexpected behavior otherwise. Feel free to ping me with any questions

api/tests/test_authentication.py
```diff
@@ -296,3 +297,27 @@ def test_session_endpoint(client):
...
+
+    rollback_email = 'test_user_signup_rollback@fake.com'
+    signup_response = app.test_client().post(
+        '/api/auth/signup/host',
+        json = {
+            'email': rollback_email,
+            'password': 'lol'
+        }
+    )
+    assert signup_response.status_code == 400
```
paulespinosa, line 314: Due to a recent change in `main` that was merged into this branch, the database does not store a user because the test does not provide a first name. Meaning, this test is no longer testing the given scenario.

paulespinosa (changes requested):

JpadillaCoding:
@tylerthome @paulespinosa @Joshua-Douglas Hey all. Thanks for your feedback! I've addressed the concerns pointed out 😄

api/openapi_server/controllers/auth_controller.py
```diff
@@ -65,6 +66,8 @@ def get_token_auth_header():
     return token
 
 def sign_up(body: dict, role: UserRole):
+    from openapi_server.controllers.admin_controller import remove_user
```
erikguntner, line 69: Question: Is importing inside a function a common practice in Python? It feels odd, but that might be from my lack of Python experience and primarily using JS.

Issue #636, opened 2023-11-04
Implement Forms API


paulespinosa:
### Overview
The API shall manage the Forms that Guests and Hosts are to fill out as Tasks that are part of the HUU housing Workflow. For the MVP, concrete Forms such as Application Forms and Profile Forms will be implemented. This issue is meant to lay the groundwork necessary for the API to manage Forms that need to be completed as Tasks that are part of the Workflow.

Forms shall have the following type of fields available:
- Text
- Date
- Phone number
- Email
- Signature: checkbox and (text box) signature 

The implementation for this issue is for the MVP. The MVP will contain hard coded Application and Profile Forms: the implementation of those Forms will populate the database with existing Form Fields for each Form.

### Action Items
Domain concerns:
- [ ] Implement Forms as Tasks. Forms are Tasks containing name, description, status, link text, link url path.
- [ ] Implement Forms as fields and responses with validation.

Endpoint concerns:
- [ ] Implement endpoint to get a Form
- [ ] Implement endpoint to save a Form
- [ ] Implement endpoint to clone a Form (For signed Forms: Guests/Hosts might be required redo a previously signed Form)

Database concerns:
- [ ] Each version of a Form has its own responses flat table. The rows of a table represent a "stack of papers" with user responses to the Form. A user can have multiple rows representing multiple submissions of the Form.
- [ ] Implement relational models to represent Form metadata: fields

### Parent User Stories
- #455 
- #533 
- #589

Joshua-Douglas:
@erikguntner, I've been inspired by your suggestions in #672 and spent some time after work creating a first-draft SQL data model that can store and retrieve user-specified forms. 

Let me know if you think it is missing anything critical. I'm planning to just hard-code the initial forms, and we can think about audit trails once we have a basic forms API working.

## Show how to define a SQL Schema that supports user defined forms

```sql
CREATE TABLE FieldTypes (
    field_type_id INTEGER PRIMARY KEY,
    type_code VARCHAR(50) NOT NULL UNIQUE,
    description TEXT,
    settings JSON,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);

INSERT INTO FieldTypes (type_code, description, settings) VALUES 
('text', 'Single line text input', '{"max_length": 255}'),
('textarea', 'Multi-line text input', '{"max_length": 1000}'),
('integer', 'Integer input', '{"min_value": 0, "max_value": 10000}'),
[...]

PR #676, opened 2024-04-16
675 Add E2E Tests to CI without Mocking

### LINKED ISSUE #675: Add Real e2e Tests to the CI Pipeline
## Overview
This issue aims to integrate end-to-end (e2e) tests with our CI pipeline to ensure all features merged have passed these tests against the real backend. The current pipeline runs our e2e test suite with backend mocking, but it does not run the test suite against our real backend. This has previously allowed features that fail these tests to be merged unintentionally. 

This change was agreed upon in the dev meeting held two weeks ago.

## Action Items
* Update the CI pipeline configuration to include running e2e tests against the real API
* Ensure that the e2e tests are configured to interact with the actual backend environment set up for testing.
* Run a test merge to verify that the e2e tests execute and block merging when tests fail.

Joshua-Douglas:
Closes #675.

  - Updated our CI pipeline to run our full e2e test suite against our real backend. This was most easily achieved using our docker container, since GitHub actions are able to manage container lifecycles when run using [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers#about-service-containers).
  - Refactored the existing pipeline steps to use our docker images. This adds some automated build checks for our images, and simplifies the CI pipeline setup.
  - Add a health check endpoint to the API, and use it to verify that the API container startup was successful. 
  - Fix minor bug in `app/cypress/e2e/create-new-password.cy.ts`. The test was short-circuiting the hostname config options.
  - bump `cryptography` to `42.0.4` since `41.0.7` has a known security vulnerability
      - `cryptography` is a transitive dependency of the `moto` library. Unfortunately the most recent versions of `moto` still use the outdated version so I had to elevate `crypography` to our `pyproject.toml` file.

It is worth noting that our API container is being used as an API and a database. In the future we might consider breaking out the database into a separate container, and populating it with some test data. I'm leaving as-is for now since our db is relatively simple at this point.

  - Our real e2e tests have broken a few times because they are a pain to run locally. This CI pipeline will prevent PR merges that break our e2e tests.

  - My biggest concern was verifying that the `test-api-real-aws` uses our real AWS service, and that `test-app-real-backend` uses the real backend. I verified both in d2e8836469f5b490ea4029951a9abd6210fd9cec by removing the AWS secret from the test config and remove the backend service from the CI pipeline. This change broke the "nomock" test suites, while the "mock" test suites continued to pass. 
      - Also, the runtimes for the nomock test suites are significantly longer

- GitHub [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers#about-service-containers) make it very easy to orchestrate multiple services 
 - The docker [HEALTHCHECK](https://docs.docker.com/reference/dockerfile/#healthcheck) command is very useful for running service containers with a lengthy startup process. GitHub action service containers will query the state of service containers using `docker inspect`. It does this query to determine when the container is ready for use. If there is no HEALTHCHECK then `docker inspect` returns a `0` exit code as soon as the container is initialized. A custom health check will ensure that `docker inspect` does not return `0` until the condition you define is met. In our case, the GH action should not progress until the API is actually available.

### Screenshots of Proposed Change
Here is the new CI pipeline flow:

image


paulespinosa (changes requested):
Lots of pros but has the downside that the runtime is longer. Find ways to utilized cache or stages.

erikguntner (commented):
This is really cool! It looks like the majority of the added time comes from building the images, so I wonder if there is a way to speed that up.

Joshua-Douglas:
I'm converting this to draft. I have some ideas on how to speed up the CI pipeline, but I'm going to implement #636 over these improvements.

PR #679, opened 2024-04-17
Check node version

### LINKED ISSUE #588: App: Node 16.x reaching end-of-life support

erikguntner:
## These changes were all taken from #638 

I ran into lots of issues while merging #638 with the main branch due to it being so outdated, so I thought it would be easier to create a new branch and add the relevant changes.

Closes #588 

  - Followed the [general solutions plan](https://github.com/hackforla/HomeUniteUs/issues/588) on the issues page, except I upgraded to Node 20.x instead of Node 18.x. 
  - I created a pre-install check to enforce a minimum node version (20.x), set in the engine's section of package.json. When a new dev installs the node packages, a script will run to check the node version first per `check-node-version.mjs`. If the minimum node version is not met, the installation will fail and a console log error will show up requiring the current node version to be satisfied.
  - Added that minimum node version is enforced and how it works in README
  
  - The docker image currently runs on the latest version of node (as of October 18th, 2023, that is Node 20.x), so there was not as much to change there. The github work flow script is set to the lts of node instead of being specified as 20.x because we are always aiming to use the lts. 
  - For enforcing the minimum node version, I set it to >=20.0.0 because Node 20.x active support ends in 11 months and security support ends in 2 years [according to this schedule](https://endoflife.date/nodejs). Therefore I chose the lowest LTS version that will give us the longest time before security support ends. 
  - As for setting `esModuleInterop` in `tsconfig.json` as "true": EsNext module does not currently have a way to support package.json exports, which was necessary for my pre-install check file `check-node-version.mjs`. To make the least amount of changes, I chose to set esModuleInterop to true and change the file extension to mjs in order for the json import to work. I also used Node's `createRequire` to mimic CommonJS "require" statement, which allows package.json imports. 
 
  - I ran the build + ran the tests + manually tested the frontend before and after the upgrade
  - I changed the node version to a ridiculous number aka `"node": ">=50.0.0" ` and ran npm i to see if I would receive the expected error
  
  - I learned a lot about modules and Typescript compiler options regarding modules. I learned that TS needs to check that: 1. files are compiled into valid output module formats, 2. ensure that imports in the outputs (aka what is compiled out to runtime) will resolve successfully, and 3. what the type of the imports are. Other ways the package.json import could have worked was to change the `moduleResolution` to `nodenext` instead as a typescript compiler option, as NodeNext makes CJS and ESM coexistable, where as simple `Node` is meant only for cjs. Since we are using ESM however, the clash of interests made the package.json import harder than expected. I also could have used ts-node but I didn't want to keep adding dependencies for only one file.

<details>
<summary>Visuals after changes are applied, testing node minimum version</summary>
  
image

</details>


paulespinosa (commented):
Consider specifying a range in `package.json`. Let's say someone is using version `22` (or 24, 26, etc) and it contains deprecations or removals of features found in version `20`. If the app (or dependencies) are using those deprecated/removed features, then that person might become confused by warnings or errors.

LTS is every even version. Consider specifying a version series in Docker such as `20-alpine` too. Version `22` will become an `LTS` version and at that time `node:lts-alpine` will use Node 22 (then 24, 26, 28, etc).

paulespinosa (approved):

PR #681, opened 2024-04-23
Retrieve first and last name when signing up with Google

### LINKED ISSUE #680: Get user's name during Google sign up
### Overview
We need to collect information about the user's name when authentication with Google.

### Action Items

- [ ] Configure AWS Cognito to collect the user's name and other relevant information.
- [ ] Save first name and last name in the database during sign up.

erikguntner:
Closes #680 

This PR allows us to collect the first and last names of users who sign up with Google. While implementing this I found some bugs while working in a "staging" environment using the real AWS user pool. Some fixes are included.

 - Updated the `/token` endpoint to only add a user to the database if they don't already exist.
  - Updated the `get_user_attr` function to get the first and last name from the user object returned from Cognito.
  - The `/confirm` route had been changed to a POST request which broke the flow we had in place, so I added back the yaml configuration and handler for the GET request. This is necessary because at the moment users confirm their profiles by clicking a link in an email that takes them to `/api/auth/confirm?email=email&code=code` which is done as a GET request. From what I've seen we can make this a POST request by using a form tag and hidden fields, but that will require refactors to the current email template. I do think this makes more sense as a POST or PUT so I left the POST implementation and just changed the name to `confirm_sign_up_v2`.
  - I also noticed session refresh wasn't working when I was getting this error:
 
Screenshot 2024-04-17 at 3 05 47 PM

  Turns out you have to use the user's username/uuid to create the secret hash when using the "REFRESH_TOKEN" auth flow instead of the email. This problem is outlined in this [GitHub issue](https://github.com/boto/boto3/issues/2736) on the boto3 issues page, and the solution came from [this stack overflow](https://stackoverflow.com/questions/54430978/unable-to-verify-secret-hash-for-client-at-refresh-token-auth/54508553#54508553) issue.
- To get that username/uuid I added the `IdToken` to the session object on sign in and then decoded it in the `refresh` controller using the `pyjwt` library. 
- Updated the `/session` endpoint to get `token` instead of the `refresh_token` property from the call to the refresh function. I noticed the object returned from refresh didn't have a `refresh_token` property on it. This makes sense since we are refreshing the access token and returning the new one to the user instead of the refresh token.
- Updated the `/user` and `/session` endpoints to return the full user object.
  - Since we've made the change to requiring the user's first, middle, and last name, we didn't have the functionality in place to do so using google authentication.
  - I manually went through the sign up and sign in flows multiple times for both google auth and email/password auth using the development and staging environments.
  - Something that still needs to be figured out is how to differentiate between signing in with Google and signing up with Google. Right now if a user sign's in with Google from our sign in page, and that user doesn't exist in AWS, a user will be created. That is an issue for us since we can't assign a user role to them. One solution I am looking into is using a pre-sign up lambda function to check if a user with that Google account already exists, and if not, throwing an error. Open to suggestions on this.

<details>

image

</details>

<details>
  
image

</details>


JpadillaCoding (changes requested):
Hey Erik! 
Took a look at your PR and noticed that the ``test_basic_auth_flow`` test is now failing. Specifically, in the portion where authentication is being tested with jwt, so perhaps it has something to do with the id_token changes?

tylerthome (approved):

PR #685, opened 2024-04-27
Loading Component Coordinator Dashboard

### LINKED ISSUE #669: Create Loader and Loading State for Coordinator Dashboard
### Overview

The current coordinator dashboard does not have loading states and needs to be updated as the Figma designs were recently changed to include these loading states. 

### Action Items

Update the loading states to render a loader based on these conditions:

- [ ] Loading 0 - 5 seconds - load state without subtitle
- [ ] Loading of 5 seconds or more - transition to load state with subtitle. 

Screen Shot 2024-03-26 at 8 49 15 PM

Screen Shot 2024-03-26 at 8 49 28 PM

johnwroge:
Closes #669 

  - Created a loading component that renders based on a timer.
  - Added the loading component and styling to the overlay option for the StyledDataGrid.
  - Updated height style on the StyledDataGrid because of the default to close when there are no rows.

   A loading spinner is needed for coordinator dashboard to act as a placeholder in the place of data based on the time it takes to fetch coordinator dashboard data from the backend. The counts in the styled tabs were not changed because they are not connected to any loading state yet. This will need to be updated when the backend is connected.

 I ran into git issues when trying to add the changes to my branch due to the linter preventing any warnings. To solve the warnings I commented out two of the unused components in the Header component and additional imports. These changes to the Header file can probably be removed in this PR and I'll bring the linter problems up in our next meeting - there may be a way to resolve this without having to comment out those components. 

I updated the Node version and merged with main which removed the comments in the Header.tsx file previously needed to commit. The tests are now passing, and I think this was caused by a local change on my device with Node or VS Code. 

  - Tests were run in the app directory and 3 failed initially in the guest invite button tests forgot password code test. After changing Node to v20 from v21, and merging with the main branch, all the tests passed. 
  - E2E tests passed with no errors in Cypress. 

<details>

before

</details>

<details>
  

after without subtitle

after with subtitle

https://github.com/hackforla/HomeUniteUs/assets/72668920/986b8522-a715-40e5-96fa-17f3379f608d

</details>


JpadillaCoding:
Hey John!
Took a look at the problem you're having with the test. 
To recreate I ran ``npm test`` in the app directory. The results I found is that main is passing all the tests on my system and I was able to get your branch to pass by pulling from main. This removes the ``container`` variable and removes the commented lines of code. 
I'm wondering why your system is showing unused imports and components while showing them as used on mine. Sometimes the linter and vscode get out of sync and vscode needs to be closed/reset- have you tried that?

johnwroge:
Hi @JpadillaCoding 

Thanks for looking into this. The issue may have had to do with vs code and the linter not being in sync, I tried restarting but didn't notice any difference. I was also using an unsupported version of node too which may have caused the testing problems. After switching to Node v.20, the errors went away. I merged with main and this removed the commented out imports and now the tests are passing now. It's odd, I'll have to look into why these linter errors are occurring on my device and not others.

app/src/views/LoadingComponent.tsx
```diff
@@ -0,0 +1,45 @@
+import {CircularProgress, Typography, Stack} from '@mui/material';
+import {useState, useEffect} from 'react';
+
+export const LoadingComponent = () => {
+  const [text, showText] = useState<boolean>(false);
```
erikguntner, line 5: I would change this to be [showText, setShowText]. It's a common pattern to do [state, setState] when destructuring a useState hook.

app/src/views/LoadingComponent.tsx
```diff
@@ -0,0 +1,45 @@
...
+        display: 'flex',
+        justifyContent: 'center',
+        alignItems: 'center',
+        height: '100%',
+      }}
+    >
+      <Stack>
+        <div style={{display: 'flex', justifyContent: 'center'}}>
+          <CircularProgress value={25} />
+        </div>
```
erikguntner, line 26: Can you get rid of this wrapping div and add the `justifyContent` property to the Stack parent component?

app/src/views/LoadingComponent.tsx
```diff
@@ -0,0 +1,45 @@
...
+              <Typography variant="h6" sx={{fontSize: 24, fontWeight: 'bold'}}>
+                Almost There!
+              </Typography>
+
+              <Typography variant="subtitle1" sx={{fontSize: 14}}>
+                Collecting key information from our database of users as quickly
+                as possible. We appreciate your patience...
+              </Typography>
+            </div>
+          </div>
```
erikguntner, line 40: Same question as above about removing the wrapping div. Then these will both be in the flex container of the parent Stack component.

erikguntner (changes requested):

app/src/views/LoadingComponent.tsx
johnwroge: Thanks Erik. I updated this.

app/src/views/LoadingComponent.tsx
johnwroge: Thanks Erik. I updated this.

tylerthome (approved):

tylerthome (approved):

PR #809, opened 2024-10-01
update run-tests with new directories

johnwroge:
  I changed the directories in the github actions run-tests-v1.yml file. 
  
  The frontend and backend test suites are failing because the directories were recently updated and the failing cases are no longer visible. The goal is to get the test suite to run correctly and then identify what is causing the failing test cases before changing the testing workflow for the new backend.


paulespinosa (approved):
Thanks John!

Issue #789, opened 2024-09-06
Convert/Port Flask code to FastAPI


lasryariel:
- Parent Issue: #771

### Overview
The goal of this task is to migrate the existing Flask codebase to FastAPI to improve performance, flexibility, and development speed. FastAPI offers features such as asynchronous request handling, Pydantic for data validation, and better dependency injection, which will enhance our current implementation. This migration will also ensure that our project uses more modern and efficient frameworks.

### Action Items
- [x] **Review Existing Code**: Identify the Flask code that needs to be ported to FastAPI. John is available for assistance in reviewing the codebase.
- [x] **Implement Pydantic and Dependency Injection**: Replace or update existing data validation mechanisms with Pydantic models. Use FastAPI's dependency injection features to streamline the service layer.
- [x] **Evaluate Conversion Benefits**: Assess which parts of the codebase would benefit from rewriting and which can be ported directly without modification.
- [x] **Implement Configuration Injection**: Ensure that configuration handling is updated to follow FastAPI's methods, including environment variable management and settings injection.
- [x] **API Error Design**: Design consistent API errors for easy consumption by the front-end. Note: It will have a basic shape with `message`, `code`, `status`. Discussed in issue #828 
  - (can be converted to it's own issue) 
- [x] **Update Alembic Migration**: Prune the old migration scripts using the instructions at [Building an Up to Date Database from Scratch](https://alembic.sqlalchemy.org/en/latest/cookbook.html#building-uptodate). Create a new Alembic migration environment.
- [x] **Create Test Users**: In development environment, pre-populate Cognito and DB with test users.
- [x] **Update Docker file configuration**: Ensure the Docker file can create a Docker image containing the FastAPI implementation and that a container can successfully run the API.
- [x] **Update GitHub Action workflows**: Update the workflows so that the new, migrated tests are run during a push to GitHub (only, not trying to deploy).
- [ ] **Flask Deprecation**: Confirm that all necessary Flask components have been successfully replaced and we are ready to deprecate Flask entirely.

### Resources/Instructions
- Focus on improving code clarity and maintainability.
- [FastAPI documentation](https://fastapi.tiangolo.com/) for reference.
- [SQLAlchemy documentation](https://docs.sqlalchemy.org/en/14/) for understanding database handling and migration steps.
- [Pydantic documentation](https://pydantic-docs.helpmanual.io/) for data validation guidance.

paulespinosa:
### Flask `auth`  and `users` endpoints
[...]

PR #811, opened 2024-10-01
Update run tests v1 github action frontend cache path

paulespinosa:
Updates the npm cache dependency path using the updated directory name.

Related #789 

  - Updated the frontend npm dependency cache path to use the updated frontend directory name.

  - The frontend code directory name was changed. This PR updates the location of the frontend code in the github action workflow. The npm cache dependency path was still using the previous frontend code location causing the action to fail.

  - Can be tested upon PR


johnwroge (approved):
Thanks Paul!

Issue #627, opened 2023-10-25
Scaffold terraform resources for incubator migration


tylerthome:
### Overview
Update the terraform resources for Home Unite Us in a fork of the Hack for LA Project Incubator repository

### Action Items

- [x] Set up resource definitions for conventional `service` module (refer to the `people-depot` definition as an example)
- [x] Add resource definition for AWS Cognito, and any additional firewall/RBAC modifications as needed (must be reviewed by ops maintainer)
- [x] Coordinate with Ops team to ensure compliance and gather general feedback for path forward

### Resources/Instructions

- [incubator repository](https://github.com/hackforla/incubator) (maintained by Ops) team

tylerthome:
Still in progress, nearing completion -- superseded by  incubator AWS terraform integration mentioned in #686

tylerthome:
This is completed as of 2024 year end - remained open due to pending PR merge on incubator repository. Ops team now plans to rewrite incubator infrastructure, and the PR for HomeUniteUs will be obsolete. Ops team has agreed to approach with a "blue/green" deployment strategy, wherein our infrastructure configuration will remain online until a suitable replacement candidate is deployed and smoke-tested by HUU team

PR #819, opened 2024-10-08
Initial incubator resources for container build

tylerthome:
Completes some items from #627 

  - add Dockerfile definitions for `base` and `app` container images to `.incubator` directory
  - add auxiliary files (e.g. nginx config, entrypoint bash script) to `.incubator` directory
  - prepare for GitHub Action deployment to incubator ECR and ECS
  - deployment to ECR
  - Dockerized nginx customization for server blocks (without TLS), in contrast to existing EC2 configuration

N/A


paulespinosa (approved):
Looks good! Renaming `docker-entrypoint.sh` will help avoid confusion. Thank you!

paulespinosa (commented):
`gunicorn` is not a dependency anymore and I can't see where it's being installed. You might be able to get away with running the `fastapi run app/main.py` command instead.

tylerthome:
> `gunicorn` is not a dependency anymore and I can't see where it's being installed. You might be able to get away with running the `fastapi run app/main.py --port 8000` command instead.

Good call @paulespinosa  - I made this change as you suggested. It's using a virtual env there now because there was an issue running either of the `fastapi` and `uvicorn`  commands as such in the entrypoint.sh. They could run with `poetry run` but this caused module references to fail while resolving from import paths. This approach worked as expected in the incubator deployment currently running at https://qa.homeunite.us, but open to any alternative approaches if this looks familiar to anyone.

paulespinosa:
> > `gunicorn` is not a dependency anymore and I can't see where it's being installed. You might be able to get away with running the `fastapi run app/main.py --port 8000` command instead.
> 
> Good call @paulespinosa - I made this change as you suggested. It's using a virtual env there now because there was an issue running either of the `fastapi` and `uvicorn` commands as such in the entrypoint.sh. They could run with `poetry run` but this caused module references to fail while resolving from import paths. This approach worked as expected in the incubator deployment currently running at https://qa.homeunite.us, but open to any alternative approaches if this looks familiar to anyone.

It looks good. Nothing off the top of my head can think of doing it any other way. Nice work!

backend/app/health.py
```diff
@@ -1,8 +1,26 @@
...
 health_router = APIRouter()
 
 
 @health_router.get("/", status_code=status.HTTP_200_OK)
 def health():
     return "UP"
+
+@health_router.get("/nginx-logs", status_code=status.HTTP_200_OK)
+def nginx_logs():
+    with open('/var/log/qa.homeunite.us/nginx-access.log') as f:
```
paulespinosa, line 13: This is a good place to introduce a `HUU_ENVIRONMENT` environment variable in `config.Settings`.

tylerthome: Ah good catch, will change this before merging

